PR #13, opened 2020-06-19
Add Wrapper for GNU ReadLine support

rocky:
Adds option `-rl` (analogous to `-clef`) to trigger GNU ReadLine support

From FriCAS's FAQ:
> 4.1: Clef does not work.  Is there an alternative?
> ...

This incorprorates the code that follow into the basic install.

Also, some basic `.gitignore`s have been started. More is desireable
though.

### DIFF
--- a/FAQ
+++ b/FAQ
@@ -117,18 +117,8 @@ then click on 'Implementations'.
 
 4.1: Clef does not work.  Is there an alternative?
 
-It is possible to use 'rlwrap' program (which is readline-based).
-The main FriCAS executable is designed to be run via line-editing
-wrapper program.  Normally for command line editing FriCAS uses
-clef.  However, it is possible to use different wrapper.
-To use 'rlwrap' do the following:
-
-1) Make a shell script to change calling convention, like:
-
-#!/bin/sh
-CFILE=$2
-shift 3
-exec /full/path/to/rlwrap -f $CFILE "$@"
+If you have GNU Readline and 'rlwrap' installed, the '-rl' option will
+use GNU Readline editing, tab completion, and history.
 
 2) Type
 
--- a/Makefile.in
+++ b/Makefile.in
@@ -142,6 +142,7 @@ install-src:
 	fi
 	echo export FRICAS >> '${COMMAND}'.tmp
 
+	$(INSTALL_SCRIPT) '$(fricas_src_srcdir)/etc/fricas-readline' '$(DESTDIR)$(bindir)/fricas-readline'
 	cat $(fricas_src_srcdir)/etc/fricas >> '${COMMAND}'.tmp
 	$(INSTALL_SCRIPT) '${COMMAND}'.tmp '${COMMAND}'
 	-rm '${COMMAND}'.tmp
--- a/src/etc/fricas
+++ b/src/etc/fricas
@@ -6,6 +6,7 @@
 #      [-ht   |-noht]       whether to use HyperDoc
 #      [-gr   |-nogr]       whether to use Graphics
 #      [-clef |-noclef]     whether to use Clef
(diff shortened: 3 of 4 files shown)


dimpase:
looks useful. As we use fricas in [sagemath](https://github.com/sagemath/sage) - a system that is certainly used a lot, this is a very useful addition, adding ease of use to FriCAS

rocky:
Followup things (some mentioned in https://groups.google.com/g/fricas-devel/c/5Bkqgp2nQl4/m/vQuiqTQAAgAJ): 

remove/alter section 4.1 of the FAQ from which this is based.
set defaulit to use GNU/Readline if `configure` detects  `rlwrap`. (This is based on the question: are there any advocates of Clef? mentioned in the above google group posting)

dimpase:
I think clef is long dead.

hemmecke:
In principle I like the contribution.
Can we leave out the .gitignore stuff from this patch?
FriCAS only recently moved to git, so it's not surprising that the .gitignore file is not yet complete.
Anyway, I would like to just maintain ONE .gitignore file in the root folder and don't clutter the other folders with .gitignore files.

 However, your patch misses one specific case, which I find important.

The fricas script usually starts sman and from that a number of other processes, but there is one exeption. If you call "fricas -nosman" then FRICASsys is called directly. Although, I don't personally care much about rlwrap, I think that one should be able to let "fricas -nosman -rl" be the same as 
`exec "rlwrap $FRICAS/bin/FRICASsys" "$@"`

Furthermore, I would like to see an error and a message, if the option -rl is given but rlwrap is not available on the system.

rocky:
Ok. Will get to this a bit later.

rocky:
@hemmecke Ok. I think everything is covered on 38f07a3 . However if something is missing (or extra) please just go ahead and adjust things. Thanks.

hemmecke:
@rocky Since I have no experience with how pull requests work in github, I've pushed a temporary branch "feature/add-rl-option" to the official repository.
I'd like to make sure that fricas-readline is **not** installed into the ${exec_prefix} directory.
I have to check but the fricas script sets the FRICAS environment variable that points to the place where the fricas files are, so
`otheropts="$otheropts -clefprog ${FRICAS}/bin/fricas-readline"`
should be fine as long as fricas-readline is in the right place.
I'll try to figure that out.

rocky:
> Since I have no experience with how pull requests work in github, I've pushed a temporary branch "feature/add-rl-option" to the official repository.

Sure - that is a reasonable thing to do. Another way to work is to pull locally from the branch and repository being merged. 

> I'll try to figure that out.

Thanks.

hemmecke:
> > Since I have no experience with how pull requests work in github, I've pushed a temporary branch "feature/add-rl-option" to the official repository.
> 
> Sure - that is a reasonable thing to do. Another way to work is to pull locally from the branch and repository being merged.

Yes I could have done this, but I would have to open a pull-request on **your** repository and your branch, if I want the additional commits happen in the public. Right?
I'm still unsure whether you get the credit (Author name in the git commit) for additional stuff that comes from me. I know I can do this locally via git rebase and squashing commits, but I've no experience with the tools that github offers.

rocky:
> Yes I could have done this, but I would have to open a pull-request on your repository and your branch, if I want the additional commits happen in the public. Right?

For that purpose, yes having a branch around is good. I thought I checked the button that allows maintainers to make additional commits off of mine, so I imagine that added some additional buttons to press. I seen this in th epast, but I don't remember where.

rocky:
Quoting from the "Allow button":

>  If checked, users with write access to fricas/fricas can add new commits to your feature/add-rl-option branch. You can always change this setting later.

So yes this might go back into my repository, but it will always be seen as a separate entity in this repository too as though you had created a separate branch.

PR #18, opened 2020-10-23
Fix efricas

s-kostyaev:
Hi!
Thanks for working on this cool project.
For now `fricas.el` doesn't work with emacs 27 and 28. I fix it in this PR. Also I add complete at point function for working with modern emacs completion (for example with `company-mode` as frontend).

### DIFF
--- a/contrib/emacs/fricas.el
+++ b/contrib/emacs/fricas.el
@@ -301,6 +301,10 @@ using \\[rename-buffer] or \\[rename-uniquely] and start a new FriCAS process.
       (fricas-run)
       (setq fricas-process (get-buffer-process (current-buffer))))
 
+    (if (> emacs-major-version 23)
+	;; enable copletion at point
+	(cl-pushnew 'fricas-complete-at-point completion-at-point-functions))
+
     (if (> emacs-major-version 21)
 	(set-process-query-on-exit-flag fricas-process nil)
       (process-kill-without-query fricas-process))
@@ -740,6 +744,32 @@ See `comint-dynamic-complete-filename'.  Returns t if successful."
                     (unless minibuffer-p
                       (message "Partially completed")))))))
     success))
+
+(defun fricas-complete-at-point ()
+  "Fricas function for complete at point."
+  (let ((start (or
+		(car (bounds-of-thing-at-point 'filename))
+		(car (bounds-of-thing-at-point 'symbol))
+		(point)))
+	(beg-of-input (fricas-beginning-of-region-pos (point))))
+    (if (and (fricas-prompt? (1- beg-of-input))
+             (save-excursion
+               (goto-char beg-of-input)
+               (looking-at " *)")))
+	(let ((files
+	       (let* ((prev-char (char-after start))
+		      (dir (if (and prev-char (char-equal prev-char ?/))
+			       (file-name-directory
+				(buffer-substring-no-properties start (point)))
+			     default-directory))
+		      (results (fricas-file-name-all-completions "" dir)))
+		 (if (not (string= dir default-directory))
+		     (mapcar (lambda (s) (concat dir s)) results)
+		   results))))
+	  (if files
+	      (list start (point) files)
+	    nil))
+      (list start (point) fricas-symbol-list))))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; yanking input
--- a/contrib/emacs/fricas.el
+++ b/contrib/emacs/fricas.el
@@ -61,7 +61,7 @@
 (require 'fricas-cpl)
 
 ;; required because of remove-if-not
-(require 'cl)
+(require 'cl-lib)
 ;; required for tab-completion
 (require 'comint)
 
@@ -194,6 +194,7 @@ This variable is buffer-local."
 (defvar fricas-stored-incomplete-input nil
   "Stored input for history cycling.")
 
+;;;###autoload
 ;;; Note that fricas-mode does *not* derive from comint mode.  A failed attempt
 ;;; can be found as fricas-comint-failed-attempt.el in the FriCAS subversion
 ;;; repository, revision 381.  comint mode assumes that the new input is always
@@ -258,8 +259,8 @@ using \\[rename-buffer] or \\[rename-uniquely] and start a new FriCAS process.
   (setq fricas-yank-file (make-temp-file "fricas" nil ".input"))
   (make-local-variable 'fricas-TeX-file)
   (setq fricas-TeX-file (make-temp-file "fricas" nil ".tex"))
-;; this is not completely right: we are going to write to the files with
-;; extension dvi, log, aux, too, so we should check for their presence.
+  ;; this is not completely right: we are going to write to the files with
+  ;; extension dvi, log, aux, too, so we should check for their presence.
   (make-local-variable 'fricas-TeX-buffer)
   (setq fricas-TeX-buffer "")          ;; buffer to put TeX code
   (make-local-variable 'fricas-save-history?)
@@ -300,7 +301,9 @@ using \\[rename-buffer] or \\[rename-uniquely] and start a new FriCAS process.
       (fricas-run)
       (setq fricas-process (get-buffer-process (current-buffer))))
 
-    (process-kill-without-query fricas-process)
+    (if (> emacs-major-version 21)
+	(set-process-query-on-exit-flag fricas-process nil)
+      (process-kill-without-query fricas-process))
     (set-process-filter fricas-process (function fricas-banner-filter))
     (process-send-string fricas-process
                          (concat ")lisp (setf |$ioHook| "
@@ -331,6 +334,7 @@ BUFFER can be either a buffer or the name of one."
   (let ((proc (get-buffer-process buffer)))
     (and proc (memq (process-status proc) '(open run stop)))))
 
+;;;###autoload
 (defun fricas ()
   "Run an inferior FriCAS process, in a BUFFER `*fricas*'.  If BUFFER exists
 but FriCAS process is not running, make new FriCAS.  If BUFFER exists and
@@ -617,7 +621,7 @@ Repeating the command at that point scrolls the list."
 (defun fricas-file-name-all-completions (pathnondir directory)
   "Returns all filenames relevant to fricas"
   (save-match-data
-    (remove-if-not
+    (cl-remove-if-not
      (function (lambda (f)
                  (or (and (string-match "\\.[^.]*\\'" f)
                           (member (match-string 0 f)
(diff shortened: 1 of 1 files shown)


hemmecke:
Thanks.
Unfortunately, I cannot review, because I do not yet have emacs > 25 and I do not use fricas.el, but rather work with https://melpa.org/#/axiom-environment (https://bitbucket.org/pdo/axiom-environment).

s-kostyaev:
Will see`axiom-environment`, thanks.

s-kostyaev:
Looks like `axiom-environment` has modes for spad and good documentation support, etc. But it looks broken for me for interactive `FriCAS` usage. It has strange indentation and doesn't support pretty-print tex output. Also `company-axiom` has some annoying behaviour - then I print `integrate(someCoolFunc(x)` and press comma and x it drops my input. Maybe it's problem with my configuration, though. I see it has better complex support for `FriCAS` development. But I think you should give `fricas.el` a try.

hemmecke:
On 10/24/20 8:35 PM, Sergey Kostyaev wrote:
> Looks like `axiom-environment` has modes for spad and good 
> documentation support, etc. But it looks broken for me for 
> interactive `FriCAS` usage.

Never experienced that. Quiet stable for me. If you detect problems,
please report to Paul Onions.

> It has strange indentation and doesn't support pretty-print tex 
> output.

Indeed, it doesn't support TeX output. But for that I now take jfricas
(https://jfricas.readthedocs.io), the Jupyter frontend for FriCAS.

There you can use ")set output tex on".

If you have a compiled version of the master branch of FriCAS, then

)set output algebra off
setFormat!(FormatMathJax)$JFriCASSupport

shows the output in some cases better. Up to now I have only realized
that the output doesn't show pretty printed, if it is too long. Then one
must resort to ")set output algebra on".

> Also `company-axiom` has some annoying behaviour - then I print
> `integrate(someCoolFunc(x)` and press comma and x it drops my input.
> Maybe it's problem with my configuration, though. I see it has better
> complex support for `FriCAS` development.

> But I think you should give `fricas.el` a try.

What do you mean by the last sentence? I used it in the beginning, but
never used tex output, because I could not read it (too small).

I've set up axiom-environment, so that I can write my session in an
.input file and then execute it step by step by just pressing Ctrl-ENTER.

hebisch:
On Fri, Oct 23, 2020 at 11:49:32AM -0700, Sergey Kostyaev wrote:
> Hi!
> Thanks for working on this cool project.
> For now `fricas.el` doesn&#39;t work with emacs 27 and 28. I fix it in this PR. Also I add complete at point function for working with modern emacs completion (for example with `company-mode` as frontend).
> 
>   https://github.com/fricas/fricas/pull/18
> -- Commit Summary --
> 
>   * fix fricas.el to work with modern emacs
>   * add complete at point function

AFAICS this patch is needed, otherwise we can not use emacs 27.
I am a little troubled that is is not clear which changes are
really needed.  Apparently just two chunks (change from cl to
cl-lib and replacement of process-kill-without-query by
set-process-query-on-exit-flag) are needed to run with emacs 27.
One chunk is needed to activate completion at point.  Some
changes seem to be pure reformatting.  Another question
is compatibility with earlier versions.  It seems that
cl-lib was added around emacs 21.  We probably do not need
to support earlier versions, but it would be good to
know exactly which versions work (and better if we can stay
compatible with earlier versions).

-- 
                              Waldek Hebisch

s-kostyaev:
> Another question
is compatibility with earlier versions.  It seems that
cl-lib was added around emacs 21.  We probably do not need
to support earlier versions, but it would be good to
know exactly which versions work (and better if we can stay
compatible with earlier versions).

http://git.savannah.gnu.org/gitweb/?p=emacs/elpa.git;a=blob;f=cl-lib.el;h=0f05c3e65efe8ccf26db4c8b554133b0696ba7c4;hb=refs/heads/externals/cl-lib
> ;; Y-Package-Requires: ((emacs "21")) ¡`emacs' package only exists in Emacs≥24!

Looks like cl-lib works with emacs 21 and newer.

s-kostyaev:
> and better if we can stay compatible with earlier versions

emacs 21 released 20 years ago. I don't thinks it's good idea to spend time for it.

hebisch:
On Thu, Mar 04, 2021 at 01:41:32PM -0800, Sergey Kostyaev wrote:
> > and better if we can stay compatible with earlier versions
> 
> emacs 21 released 20 years ago. I don't thinks it's good idea to spend time for it.
> 

If support is a simple as something like:

(if (> emacs-major-version 20)
    (require 'cl-lib)
    (require 'cl))

then why not.  If support for old emacs would require more work,
then sure, it is not worth time spent.

BTW: Main thing is what is really needed from the patch?  The

;;;###autoload

things looks like junk, are they doing anything useful?  And the
change from 'remove-if-not' to 'cl-remove-if-not' (and few similar
later), is this needed?

Some changes seem to just change indentation of comments.  AFAICS
current version is consistently indented, so even if somebody
likes different indentation rules it makes sense to keep
indentaion as is.

BTW2:

+    (if (> emacs-major-version 23)
+       ;; enable copletion at point
                  ^^^^^^^^^
                  completion
+       (cl-pushnew 'fricas-complete-at-point completion-at-point-functions))

-- 
                              Waldek Hebisch

s-kostyaev:
> On Thu, Mar 04, 2021 at 01:41:32PM -0800, Sergey Kostyaev wrote: > and better if we can stay compatible with earlier versions emacs 21 released 20 years ago. I don't thinks it's good idea to spend time for it.
> If support is a simple as something like: (if (> emacs-major-version 20) (require 'cl-lib) (require 'cl)) then why not. 

Kind of. But on all function call we also need this kind of `if`. It will increase cost of support in the future.

> The ;;;###autoload things looks like junk, are they doing anything useful?

Sure. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html

> And the change from 'remove-if-not' to 'cl-remove-if-not' (and few similar later), is this needed? 

this are functions from obsolete `'cl` library.

> Some changes seem to just change indentation of comments.

Just modern code style. Can be dropped.

> AFAICS current version is consistently indented, so even if somebody likes different indentation rules it makes sense to keep indentaion as is. 

Sounds reasonable.

> BTW2: + (if (> emacs-major-version 23) + ;; enable copletion at point ^^^^^^^^^ completion + (cl-pushnew 'fricas-complete-at-point completion-at-point-functions))

Will fix it. Thanks for the review.

hebisch:
On Thu, Mar 04, 2021 at 02:48:59PM -0800, Sergey Kostyaev wrote:
> > On Thu, Mar 04, 2021 at 01:41:32PM -0800, Sergey Kostyaev wrote: > and better if we can stay compatible with earlier versions emacs 21 released 20 years ago. I don't thinks it's good idea to spend time for it.
> > If support is a simple as something like: (if (> emacs-major-version 20) (require 'cl-lib) (require 'cl)) then why not. 
> 
> Kind of. But on all function call we also need this kind of `if`. It will increase cost of support in the future.

Clearly we do not want to put conditions on calls, that would be
too much.  But see question below.
 
> > The ;;;###autoload things looks like junk, are they doing anything useful?
> 
> Sure. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html

Yes, I could guess something like that.  However, AFAICS efricas to
do anything needs to load 'fricas.el' and once this is loaded
I do not see why autoload cookies would do anything useful.
And efricas seem to work fine without any autoload cookies in
'fricas.el'.  So the question is if you have some other application
that needs autoload cookies?  If yes, then we should add
comment so that next person modifying 'fricas.el' keeps them.
Otherwise we should not add them.

> > And the change from 'remove-if-not' to 'cl-remove-if-not' (and few similar later), is this needed? 
> 
> this are functions from obsolete `'cl` library.

I did a small test and just using (require 'cl-lib') and
'set-process-query-on-exit-flag' change works in emacs 27.
This was rather small test so maybe I missed something.
Or do you mean that old names work in emacs 27 but are going
to vanish later?

-- 
                              Waldek Hebisch

s-kostyaev:
> > The ;;;###autoload things looks like junk, are they doing anything useful? Sure. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html
>
> Yes, I could guess something like that. However, AFAICS efricas to do anything needs to load 'fricas.el' and once this is loaded I do not see why autoload cookies would do anything useful. And efricas seem to work fine without any autoload cookies in 'fricas.el'. So the question is if you have some other application that needs autoload cookies? If yes, then we should add comment so that next person modifying 'fricas.el' keeps them. Otherwise we should not add them.

It will work without autoload cookies. But with this magic comment we don't need to `(require 'efricas)` before use. This can be very helpful for decreasing emacs init time. All modern emacs packages has autoloads and it is [good practice](https://github.com/bbatsov/emacs-lisp-style-guide#loading-and-autoloading) for elisp packages.

> > And the change from 'remove-if-not' to 'cl-remove-if-not' (and few similar later), is this needed? this are functions from obsolete `'cl` library.
>
> I did a small test and just using (require 'cl-lib') and 'set-process-query-on-exit-flag' change works in emacs 27. This was rather small test so maybe I missed something. Or do you mean that old names work in emacs 27 but are going to vanish later?

`C-h f remove-if-not RET`
>   This function is obsolete since 27.1;
  use ‘cl-remove-if-not’ instead.

And so on and so forth. Even if it is not deleted right now, it can be deleted in the future. See `process-kill-without-query`.

hebisch:
On Fri, Mar 05, 2021 at 01:56:02AM -0800, Sergey Kostyaev wrote:
> > > The ;;;###autoload things looks like junk, are they doing anything useful? Sure. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Autoload.html
> >
> > Yes, I could guess something like that. However, AFAICS efricas to do anything needs to load 'fricas.el' and once this is loaded I do not see why autoload cookies would do anything useful. And efricas seem to work fine without any autoload cookies in 'fricas.el'. So the question is if you have some other application that needs autoload cookies? If yes, then we should add comment so that next person modifying 'fricas.el' keeps them. Otherwise we should not add them.
> 
> It will work without autoload cookies. But with this magic comment we don't need to `(require 'efricas)` before use. This can be very helpful for decreasing emacs init time. All modern emacs packages has autoloads and it is [good practice](https://github.com/bbatsov/emacs-lisp-style-guide#loading-and-autoloading) for elisp packages.

Sorry, can you give simple Yes/No answer to the following:

1) Without special action Emacs will not look at the file and will
   not know about autoload cookies.  Yes or no?
2) 'efricas' to start fricas needs to load 'fricas.el'.  Yes or no?
3) Do you have other use for 'fricas.el' than using it via
   'efricas'.  Yes or no?

If answer to first two is yes and to third is no, then clearly
cookies that you put are dead code and good practice is to
remove dead code.

> > > And the change from 'remove-if-not' to 'cl-remove-if-not' (and few similar later), is this needed? this are functions from obsolete `'cl` library.
> >
> > I did a small test and just using (require 'cl-lib') and 'set-process-query-on-exit-flag' change works in emacs 27. This was rather small test so maybe I missed something. Or do you mean that old names work in emacs 27 but are going to vanish later?
> 
> `C-h f remove-if-not RET`
> >   This function is obsolete since 27.1;
>   use ???cl-remove-if-not??? instead.
> 
> And so on and so forth. Even if it is not deleted right now, it can be deleted in the future.

OK, so let us forget about compatibility with versions before 21 and
go with 'cl-' variants.

-- 
                              Waldek Hebisch

PR #21, opened 2020-10-29
doc: book: add missing backslash to Tex command

nsajko:
A spadtype Tex command was missing its backslash.


hemmecke:
Fixed together with some other places that did not display -> correctly.

PR #23, opened 2020-10-29
doc: awk regexp escaping fixes

nsajko:
Add missing backslashes, remove superfluous backslashes.

### DIFF
--- a/src/doc/ht.awk
+++ b/src/doc/ht.awk
@@ -185,7 +185,7 @@ inXmpLines==1 {
         gsub(/\\bs\\{\\}/, "\\bs{}")
         gsub(/ /, "\\ ")
         gsub(/&/, "\\\\&")
-        gsub(/\#/, "\\#")
+        gsub(/#/, "\\#")
         gsub(/\$/, "\\$")
         gsub(/%/, "\\%")
         gsub(/_/, "\\_")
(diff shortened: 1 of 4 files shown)


hemmecke:
Have you tried to compile?
Maybe you know better about escaping # % and _ in awk, but I guess that keeping the backslash would not hurt.

I am worried a bit about an obviously missing double backslash here
/^-- \end{noOutputXtc}$/. How on earth could that have worked to produce book.pdf? I must look deeper.

nsajko:
> Have you tried to compile?

Yes, and I didn't notice any regressions.

> Maybe you know better about escaping # % and _ in awk, but I guess that keeping the backslash would not hurt.

# % _ are not special characters in regular expressions. This means that escaping them is undesirable: in practice it probably doesn't hurt, but gawk produces annoying warnings during run time in such cases, and the semantics are actually undefined according to POSIX.

hebisch:
On Thu, Oct 29, 2020 at 01:13:55AM -0700, Neven Sajko wrote:
> Add missing backslashes, remove superfluous backslashes.
> 
>   https://github.com/fricas/fricas/pull/23
> 
> -- Commit Summary --
> 
>   * doc: awk regexp escaping fixes
> 
> -- File Changes --
> 
>     M src/doc/ht.awk (4)
>     M src/doc/htex2input.awk (6)
>     M src/doc/spool2tex.awk (4)
>     M src/doc/ugepsf.awk (2)
> 
> -- Patch Links --
> 
> https://github.com/fricas/fricas/pull/23.patch
> https://github.com/fricas/fricas/pull/23.diff

That looks OK, I think we should merge it.

-- 
                              Waldek Hebisch

hemmecke:
Indeed. I've merged it.

PR #28, opened 2021-01-07
a little set of typos

fchapoton:
found with codespell

There remains many others, of course.


oldk1331:
Hi, one small problem: line 583 of  src/input/test.input should not contain "x".

PR #31, opened 2021-01-20
Update plot3d.spad

fchapoton:
remove /em that does not display correctly on https://fricas.github.io/api/Plot3D.html

### DIFF
--- a/src/algebra/plot3d.spad
+++ b/src/algebra/plot3d.spad
@@ -32,13 +32,13 @@ Plot3D() : Exports == Implementation where
   Exports ==> PlottableSpaceCurveCategory with
 
     pointPlot : (F -> P, R) -> %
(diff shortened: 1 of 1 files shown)


hemmecke:
That seems to be an obvious bug. It somehow seems that /em should have been \em. But then I would also remove the braces {\em ...}, don't you agree?
I have not yet gone over all such cases, because it is still not clear in what markup the ++ docstrings should be written.
However, you are right. "/em" is a bug.

fchapoton:
So what should I do ?

hemmecke:
If you also remove the braces, I'll merge it. But for the moment I do not think it is wise to invest time in formatting ++ docstrings until it is clear in what format they should be written (and that may take as long as HyperDoc exists.)
Nevertheless, thank you for all your other corrections.

fchapoton:
I did remove the braces.

hemmecke:
squashed and merged

PR #44, opened 2021-03-08
improve the README a little bit

stylewarning:


hebisch:
There is now updated README

PR #45, opened 2021-03-08
Clean up and restructure INSTALL file

stylewarning:
This is an attempt to make `INSTALL` a little nicer and more straightforward. It puts the essential instructions on top, while putting a series of "appendices" after.

I did go ahead and delete some things, like the paragraphs about Spectre mitigations, and a modicum of information about GCL.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -1,115 +1,230 @@
-Quick installation:
+PREREQUISITES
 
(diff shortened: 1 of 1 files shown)


hebisch:
There is now updated INSTALL

PR #46, opened 2021-03-08
modernize package definitions and move host Lisp config to separate file

stylewarning:
This commit modernizes the package definitions, and additionally
moves host-Lisp-specific configuration into its own file.

Tested by building FriCAS out of tree on Linux. My only fear is that even though the changes I made are reasonable and perfectly ANSI-compliant, they still might not work on GCL...


hebisch:
On Sun, Mar 07, 2021 at 11:11:02PM -0800, Robert Smith wrote:
> This commit modernizes the package definitions, and additionally
> moves host-Lisp-specific configuration into its own file.

That is largish patch and I had only superfical look at it.
However, I few problems with it:

- 2 space indent.  Maybe you manage it better, but to see
  code structure I need at least 4 space indent.
- more files => more complicated Makefile
- massive use of uninterned symbols when strings or normal
  symbols work fine.

-- 
                              Waldek Hebisch

stylewarning:
- I'm not indenting with respect to an N-space indent rule. I'm indenting with respect to a "canonical indentation scheme" (for which two spaces is the standard for top-level body forms). This style of indentation is more-or-less standard in the Common Lisp and Emacs Lisp worlds. Going against this would mean going against the default configuration of SLIME/SLY, Emacs, LispWorks, Clozure CL IDE, and Allegro CL. I would strongly petition individuals to use the default indentation. (Also note that `&REST` and `&BODY` forms are indented differently; keyword arguments are aligned; etc.)
- The Makefile structure is only marginally impacted, and I think that could be simplified as well. I have only thought about it a little bit today, so I don't a direct suggestion right now. What I can say is that, in an ideal scenario, one should not fear adding new files to a project. If there is fear, it usually means there's a deficiency in the build specification.
- The uninterned symbols are a practice I adopted so as to work better with other Lisp compilers, like Allegro, which have a "modern mode". Moreover, these symbols are garbage collected since they're not spuriously interned into the `*PACKAGE*` (which causes them to show up with e.g. `apropos`). I'm happy to revert to using capitalized strings if this project prefers it.

My overarching goal with this and similar patches are to increase the organization and "declarative" programming style of the Lisp system, while still respecting the dominant paradigm which the project follows. (For example, I will not be submitting any patches to, say, convert all iteration to tail-recursion.) Of course, it's not just for the sake of it, it's primarily to:

- Make debugging easier, since implementation-specific code is cordoned off, and there are fewer instances of "image modification".
- Make the code easier to understand for newer and experienced developers alike.

I hope we can agree that a `defpackage` form with exported symbols is easier to read than a `make-package` form in one part of the file, and various `export`s elsewhere. :)

PR #48, opened 2021-03-08
de-macroify much of fricas-lisp.lisp

stylewarning:
This commit removes lots of DEFMACROs and replaces them with DEFUNs
when possible. In general, this is preferable for two reasons:

(1) The functions are still available for use as high-order functions.

(2) Functions avoid multiple evaluation of the arguments.

(3) In many cases, the code is simplified.

(4) The assembly code can be readily inspected for performance
    analysis.

This commit also:

- Re-orders some of the definitions so the inline expansions catch.

- Deletes some old SBCL code for constructing specialized vectors.

### DIFF
--- a/src/lisp/fricas-lisp.lisp
+++ b/src/lisp/fricas-lisp.lisp
@@ -2,6 +2,7 @@
 ;;; differences between Lisp dialects.
 
 (in-package "FRICAS-LISP")
+
 #+:cmu
 (progn
      (defvar *saved-terminal-io* *terminal-io*)
@@ -200,12 +201,12 @@ with this hack and will try to convince the GCL crowd to fix this.
 )
 
 (defun save-core (core-image)
-     (save-core-restart core-image nil))
+  (save-core-restart core-image nil))
 
 ;; Load Lisp files (any LOADable file), given as a list of file names.
 ;; The file names are strings, as appropriate for LOAD.
 (defun load-lisp-files (files)
-  (mapcar #'(lambda (f) (load f)) files))
+  (mapc #'load files))
 
 ;;; How to exit Lisp process
 #+:GCL
@@ -218,8 +219,8 @@ with this hack and will try to convince the GCL crowd to fix this.
 
 #+:sbcl
 (defun exit-with-status (s)
-    (setf *terminal-io* *saved-terminal-io*)
-    (sb-ext::quit :UNIX-STATUS s))
+  (setf *terminal-io* *saved-terminal-io*)
+  (sb-ext:quit :UNIX-STATUS s))
 
 #+:clisp
 (defun exit-with-status (s) (ext::quit s))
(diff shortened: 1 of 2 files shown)


stylewarning:
(I think this PR in its current state crowbars modular arithmetic for some reason. Will have to investigate.)

oldk1331:
I canceled the CI runs so that logs will be available.

I should add a timeout for those CI runs.

stylewarning:
Tests now pass. I have not rebased on top of @hebisch's latest changes on master, because I recommend merging #51, so I can rebase on top of that instead. Performance hasn't been substantially impacted, from what I can tell by benchmarking the test suite. (It's within 5 seconds of one another in a single measurement.)

stylewarning:
I decided to rebase anyway; I figure it's easier to look at and review if it's truly on top of current master.

hebisch:
On Sun, Mar 07, 2021 at 11:48:11PM -0800, Robert Smith wrote:
> This commit removes lots of DEFMACROs and replaces
> them with DEFUNs when possible. In general, this is
> preferable for two reasons:

I our case those to reasons are not good

> (1) The functions are still available for use as
>     high-order functions.

Macros are mostly used in situation when we do not need to
call them as higher order functions.

There is also another aspect: higher order use usually
prevents inlining.  Using performace critical
function in way which can not be inlined is a
performance bug.  Macros prevent such problem.

> (2) Functions avoid multiple evaluation of the
>     arguments.

This should not be a problem: if needed macros (like
SPADCALL) use LET binding to avoid unwanted exaliation.
When argument is used only once, there are no danger
of multiple evaluation.

There is one good reason to eliminate macros: macros
increase complexity.  So things that do not need to
be macros should not.  However, before any massive
change we should do proper research.  Namely, main
reason to use macros is for optimization.  Most
macros when introduced gave quite measurable
speedups for corresponding operations.  So we
need to know:

- do compilers really respect inline directive (I hope
  so, but hope in not enough)?
- do inlining give at least as good code as macros?
  Significant effect of our macros is from inserting
  explicit or implicit declarations, is this effect
  preserved with inline functions?
- what is effect on compile time?

I would hope that inline functions give us essentially
the same object code as macros.  But I have seen
strange irregularities in compilers, so for me it
is thing to check.

In mention compile time because some output of
Spad compiler triggers pathological behaviour
of sbcl compiler (long compile time and high
memory use).  IMO it important to know if
sbcl behaviour changes when we replace macros
by inline functions.

Let me say that ATM I would expect that change
from macros to inline functions will be rather
neutral. Namely we will replace one kind of bugs
(unwanted evaluation), by another (failing to
inline perfomance critical code).  And performace
of correct code probably will be the same.

Now, taking about removing macros I am interested
in finding macros that have no good reason to
be macros and could be replaced by ordinary
functions with no or negligeble degradation in
performace.  Or simplifying code so that
indirection via macros is not needed (but
that probably would require going at least
to Boot level).

-- 
                              Waldek Hebisch

stylewarning:
Inlined functions are well-respected on implementations like SBCL, but I have no idea about goofball implementations like GCL.

PR #51, opened 2021-03-08
fix #\Return stripping code

stylewarning:
The old code used EQ which is not guaranteed to
work for characters. In the process we also simplify
the code a bit.

This is a fix for @hebisch's recent change to master.

### DIFF
--- a/src/boot/initial-env.lisp
+++ b/src/boot/initial-env.lisp
@@ -87,13 +87,13 @@
 
 (defun |shoeread-line| (st)
   (let ((str (read-line st nil nil)))
-      (if (not str) str
-          ;; remove dos CR
-          (let ((l_str (length str)))
-              (if (and (> l_str 0)
-                       (eq (char str (1- l_str)) #\Return))
-                  (subseq str 0 (1- l_str))
-                  str)))))
+    (if (null str)
+        str
+        (let ((len (length str)))
+          (if (or (zerop len)
+                  (char/= #\Return (char str (1- len))))
+              str
+              (subseq str 0 (1- len))))))) ; Trim CR from end.
 
 (defun |shoePLACEP| (item)
   (eq item nil))
--- a/src/interp/macros.lisp
+++ b/src/interp/macros.lisp
@@ -582,57 +582,45 @@
 
 ;; moved here from preparse.lisp
 
-(defun NEXT-TAB-LOC (i) (* (1+ (truncate i 8)) 8))
+(defvar tab-size-in-spaces 8
+  "How many spaces do we consider a #\Tab character?")
 
-(defun INDENT-POS (STR)
-  (do ((i 0 (1+ i))
-       (pos 0))
-      ((>= i (length str)) nil)
-      (case (char str i)
-            (#\space (incf pos))
-            (#\tab (setq pos (next-tab-loc pos)))
-            (otherwise (return pos)))))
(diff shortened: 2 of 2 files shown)


oldk1331:
This works, I tested it in my repo.  But if we set them as "text", then it is not show in the "Language statistics bar".

hebisch:
On Mon, Mar 08, 2021 at 09:00:53AM -0800, Robert Smith wrote:
> The old code used EQ which is not guaranteed to
> work for characters. In the process we also simplify
> the code a bit.

First, sorry for delay, you raise a lot of issues at it takes
time to look at them.  Thanks for catching EQ, we definitely
want to fix such things.  Concerning your patch, I feel
unconfortable applying 'length' to 'nil'.  Yes, it is good
Lisp, but in FriCAS a lot of things have declared types
and mixing types in this way is a potential trap for
future persons working with this code.  Comment helps,
but explicit test for nil seem clearer.

Next thing: I am not sure if you looked at messages at
fricas mailing list.  There was a little discussion of
this code, and in particular its handling of #\Return
mimics what is done in 'expand-tabs' in
'src/interp/macros.lisp'.  So, it is natural to fix
also EQ issue in 'expand-tabs' in the same commit.

BTW. There is nonzero chance that EQ issue is repeated
in some other place.  Searching for such thing is
somewhat tricky, so it is fine if you do not try.

-- 
                              Waldek Hebisch

hebisch:
On Tue, Mar 09, 2021 at 11:32:03PM -0800, Robert Smith wrote:
> @hebisch I made your suggestions in two separate commits. I had two questions:
> 
> 1. I couldn't find anywhere in the source code where `expand-tabs` was being called. Am I missing something?

incl.boot:278:            str  :=  EXPAND_-TABS first ss

Boot is case-sensitive and '_' has double duty: it works as escape
for characters that otherwise would be illegal in Boot/Spad names.
'_' that is not an escape is just includend in the name.  Exact
rules a bit more complicated than what I wrote above, but in
Boot/Spad you can make almost arbitrary name by escaping characters.

In new code I use convention that Lisp functions called from
Boot/Spad should have the same name (including capitalization)
in both Lisp and Boot/Spad.  To avoid tons of silly wrappers
standard CL functions are called from Boot/Spad using escapes.

There are still several functions in our hand-written Lisp
(like expand-tabs) that need escapes for calls...

> 2. I rewrote `expand-tabs` significantly, which I suspect you don't like in general. However, in addition to using the wrong equality predicate, it used `indent-pos` which was quadratic in time, and `expand-tabs` itself was quadratic in space. So you have this cubic time and quadratic space thing for something that can be done in linear time and space. But even more than that, I allowed it to properly recognize `#\Newline` and compute the tabs correctly in that case.

Well, 'expand-tabs' always is called on lines, so it will never see
a #\Newline.  Expected use case is no tabs at all (our codebase
is supposed to contain no tabs in Boot/C/Spad/Lisp files).
More speed is nice, but other things in Spad compiler take _much_
more time (I use profiler it identify hot spots and expand-tabs and
indent-pos were low in list of things to improve).

We are heading towards release, so it is time to check that our
code works on all supported Lisps.  Your patch may add features
not present in our code (I am not sure, but it is written differently),
so I will run few tests with your patch included.  So please
wait a little.

-- 
                              Waldek Hebisch

PR #53, opened 2021-03-08
specify boot/spad as text and lisp/clisp as Common Lisp

stylewarning:
GitHub currently gravely misclassifies the contents of the repository. This commit ensures `.lisp` and `.clisp` are seen as Common Lisp files, and `.boot` and `.spad` are seen as text files.

GitHub obviously doesn't recognize `.boot` and `.spad` as languages on their own, and that would require changes to a Ruby project called "Linguist". Linguist also doesn't have support for "custom" or "unknown" languages, so we make a best effort compromise and just list those files as text files.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,7 @@
 * text=auto eol=lf
+
+*.boot linguist-language=Text
+*.spad linguist-language=Text
+*.lisp linguist-language=Common-Lisp
+*.clisp linguist-language=Common-Lisp
+


oldk1331:
This works, I tested it in my repo. But if we set them as "text", then it is not show in the "Language statistics bar".

stylewarning:
> This works, I tested it in my repo. But if we set them as "text", then it is not show in the "Language statistics bar".

Another option is to just approximate Boot and Spad as “Algol” or some such. What do you think about that?

In any case, either option is better than having things classified as Clojure.

hebisch:
On Tue, Mar 09, 2021 at 01:34:07PM -0800, Robert Smith wrote:
> > This works, I tested it in my repo. But if we set them as "text", then it is not show in the "Language statistics bar".
> 
> Another option is to just approximate Boot and Spad as ???Algol??? or some such. What do you think about that?
> 
> In any case, either option is better than having things classified as Clojure. 
I must admit that I do not understand what the problem really is.
If I read correctly the first message, Github has buggy tool
that guesses language of files.  And they do not offer any
way for us to fix this.  But what are consequences of their
bug?   If this is just wrong stats, then I would say that
this is Github problem, not our problem.  In particular
I would object putting wrong (fake) tag.  OTOH if there is some
concrete problem (say avoiding mangled views) that can
be solved by fake tag, then we may consider it.

Just another thougt: tagging code as text may give quite
wrong impression about our code size.

BTW: Of course, I support correctly tagging Commmon Lisp.

BTW2: Github tool looks like a bad joke.  It is tempting to
do better joke, say claim that our code is something ridiculous
like BrainFack or Cobol (if they support that).

-- 
                              Waldek Hebisch

stylewarning:
> On Tue, Mar 09, 2021 at 01:34:07PM -0800, Robert Smith wrote: > This works, I tested it in my repo. But if we set them as "text", then it is not show in the "Language statistics bar". Another option is to just approximate Boot and Spad as ???Algol??? or some such. What do you think about that? In any case, either option is better than having things classified as Clojure.
> I must admit that I do not understand what the problem really is. If I read correctly the first message, Github has buggy tool that guesses language of files. And they do not offer any way for us to fix this. But what are consequences of their bug? If this is just wrong stats, then I would say that this is Github problem, not our problem. In particular I would object putting wrong (fake) tag. OTOH if there is some concrete problem (say avoiding mangled views) that can be solved by fake tag, then we may consider it. Just another thougt: tagging code as text may give quite wrong impression about our code size. BTW: Of course, I support correctly tagging Commmon Lisp. BTW2: Github tool looks like a bad joke. It is tempting to do better joke, say claim that our code is something ridiculous like BrainFack or Cobol (if they support that).
> […](#)

GitHub uses a tool called “Linguist” to heuristically determine the language each file is written in, in order to both report what languages the project is written in (so it may be searched and displayed), and to display highlighting of the files I believe.

GitHub is currently grossly misrepresenting the repository by assuming for example that all Boot files are actually Clojure files.

GitHub and Linguist offer no way to configure a repository to report on languages arbitrarily; one must select from a fixed set. That set obviously does not include Boot or Spad. They also don’t offer a “Custom” or “Other” language.

So the options are:

- Claim them as text, in which case the relative percentages of code will be misrepresented, but at least no language is misrepresented. (this PR)
- Choose a language that closely resembles, like Algol (lol), but understand the project will appear as if it were written in Algol to passing eyes.
- Do nothing, and stick with the hodgepodge of languages GitHub thinks this project is, like Clojure.

hemmecke:
> So the options are:
> 
> - Claim them as text, in which case the relative percentages of code will be misrepresented, but at least no language is misrepresented. (this PR)
> - Choose a language that closely resembles, like Algol (lol), but understand the project will appear as if it were written in Algol to passing eyes.
> - Do nothing, and stick with the hodgepodge of languages GitHub thinks this project is, like Clojure.

Axtually none of these look appealing. What about filing an issue to the
github team that they think about the problem of misrepresented
projects. Maybe they don't care, because they just want get the
mainstream projects correct. But maybe, they at least allow for "other
language" as a tag. Who knows?

Ralf

stylewarning:
> So the options are: - Claim them as text, in which case the relative percentages of code will be misrepresented, but at least no language is misrepresented. (this PR) - Choose a language that closely resembles, like Algol (lol), but understand the project will appear as if it were written in Algol to passing eyes. - Do nothing, and stick with the hodgepodge of languages GitHub thinks this project is, like Clojure.
> Axtually none of these look appealing. What about filing an issue to the github team that they think about the problem of misrepresented projects. Maybe they don't care, because they just want get the mainstream projects correct. But maybe, they at least allow for "other language" as a tag. Who knows? Ralf

This issue has been filed [here](https://github.com/github/linguist/issues/3665) and has been open since 2017. Interestingly, it's still an active issue, and also interestingly, as of 3 days ago, the top suggestion is to specify the language as `Text` if it's not supported.

oldk1331:
How about we treat both boot/spad files as Common Lisp?  It's not perfect, but it's better than shown as Clojure or not shown at all.

stylewarning:
I'll close this PR. You all know where `.gitattributes` is now, and the syntax for making this change. Decide and implement amongst yourselves what you see best for this project.

PR #58, opened 2021-04-07
.gitattributes: boot files are not Clojure

vindarel:
Hi, discovering FriCAS, a fork of Axiom, was great, but seeing it was composed mainly of Clojure code, as GitHub says, was puzzling. It turns out .boot files are counted as Clojure ([example](https://github.com/fricas/fricas/blob/master/src/interp/msg.boot)), but I am pretty sure they are not. 

I edited the .gitattributes to teach GitHub ([doc](https://github.com/github/linguist/blob/master/docs/overrides.md)). However, are they Spad files?

With this PR, the language count becomes: 41% C, 30% PostScript, 10% Common Lisp etc instead of 33% Clojure.

(Shall we go further and count .spad files as Lisp??)

Regards,

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,5 @@
 * text=auto eol=lf
+
+# Tell GitHub that .boot files are not Clojure.
+*.boot linguist-language=Spad 
+*.spad linguist-language=Spad


hemmecke:
Thanks, but anything we change will still be wrong. It is a problem of github that we cannot much do about.
Declaring .boot or .spad to be LISP is as wrong as declaring them clojure files.
https://github.com/fricas/fricas/pull/53

vindarel:
>  Declaring .boot or .spad to be LISP is as wrong as declaring them clojure files.

Sure :)

Declaring them as Spad and Boot might be the way to go?

vindarel:
Just saw and read the related issue. The previous PR was good to me. There's very wrong and less wrong…

hemmecke:
> Declaring them as Spad and Boot might be the way to go?
For my taste, declaring them to be SPAD and BOOT would make sense, if they then are counted by linguist under the "Other" percentage.

dimpase:
one can try to do a PR for github like this one: https://github.com/github/linguist/issues/3462

oldk1331:
Closed by #100.

PR #68, opened 2021-11-20
Make HyperDoc scrolling on a touchpad more robust.

pdo:


oldk1331:
I have tested this myself, indeed works as expected: now touch pad scrolling on "command examples" will not be interpreted as click.

PR #73, opened 2021-12-17
fricas-lisp: add support for C types char, short, long for SBCL

nsajko:
fricas-foreign-call extension for the SBCL case.

### DIFF
--- a/src/lisp/fricas-lisp.lisp
+++ b/src/lisp/fricas-lisp.lisp
@@ -463,7 +463,10 @@ with this hack and will try to convince the GCL crowd to fix this.
 (eval-when (:compile-toplevel :load-toplevel :execute)
 
 (setf *c-type-to-ffi* '(
+    (char SB-ALIEN::char)
+    (short SB-ALIEN::short)
     (int SB-ALIEN::int)
+    (long SB-ALIEN::long)
     (c-string SB-ALIEN::c-string)
     (double SB-ALIEN::double)
 ))


nsajko:
ping @hebisch

oldk1331:
@nsajko  May I ask the reason behind this patch?  Clearly FriCAS doesn't use these types, so I presume you want to use these types in your own programs?

hebisch:
On Fri, Dec 17, 2021 at 01:41:01AM -0800, Neven Sajko wrote:
> fricas-foreign-call extension for the SBCL case.
> 
>   https://github.com/fricas/fricas/pull/73
> 
> -- Commit Summary --
> 
>   * fricas-lisp: add support for C types char, short, long for SBCL
> 
> -- File Changes --
> 
>     M src/lisp/fricas-lisp.lisp (3)
> 
> -- Patch Links --
> 
> https://github.com/fricas/fricas/pull/73.patch
> https://github.com/fricas/fricas/pull/73.diff

Sorry for long delay.  I think that we should keep support
for various Lisp flavours in sync, that is have also support
for other Lisp variants.  Another thing is 'char' and 'short'.
At binary level they should work if you declare them as 'int'.
OTOH if one wants to follow C distinctions, then much more
is needed.


-- 
                              Waldek Hebisch

nsajko:
> May I ask the reason behind this patch? Clearly FriCAS doesn't use these types, so I presume you want to use these types in your own programs?

@oldk1331 At one point I had some Fricas code that was too slow, so I reimplemented it as a C++ function with 64-bit integer parameters (`long`). I needed this proposed change to be able to use my C++ code from the rest of my Fricas code. Excuse the delayed answer.

The whole thing was just a bit of fun: https://gitlab.com/nsajko/fricas_guessing

> OTOH if one wants to follow C distinctions, then much more is needed.

@hebisch just an idea, but perhaps it would be simpler to provide just the explicitly-sized types, 8-bit, 16-bit, 32-bit, 64-bit integers; and then remove the old-style C types like `char`, `short`, `int` and `long`.

PR #86, opened 2022-06-11
Detect buggy sbcl versions on 32-bit platforms

oldk1331:
As discussed in https://github.com/fricas/fricas/issues/55, this patch checks for buggy sbcl versions on 32-bit platforms, to prevent people from falling into this issue again.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,8 @@
+2022-06-11  Qian Yun  <oldk1331@gmail.com>
+
+	* configure.ac: Detect buggy sbcl versions on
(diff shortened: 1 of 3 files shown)


oldk1331:
I have tested this code on a debian 11.3 system.

hebisch:
On Fri, Jun 10, 2022 at 08:40:41PM -0700, Qian Yun wrote:
> I have tested this code on a debian 11.3 system.

Looks OK, please commit.

-- 
                              Waldek Hebisch

PR #89, opened 2022-06-11
Fix environment variable on macOS to make dmg package work out of box

oldk1331:
This patch correctly set environment variable, so that we can start fricas from terminal and from "double click app icon", without hack.

### DIFF
--- a/contrib/macos/FriCAS.c
+++ b/contrib/macos/FriCAS.c
@@ -22,10 +22,12 @@ int main(int argc, const char *argv[]) {
   CFStringAppend(path, prefix);
   CFStringAppendCString(path, "/", encoding);
   CFStringAppend(path, resources);
+  CFStringAppendCString(path, "/usr/local/", encoding);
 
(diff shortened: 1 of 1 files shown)


hebisch:
On Sat, Jun 11, 2022 at 12:57:57AM -0700, Qian Yun wrote:
> This patch correctly set environment variable, so that we can start fricas from terminal and from &quot;double click app icon&quot;, without hack.
> 
>   https://github.com/fricas/fricas/pull/89
> 
> -- Commit Summary --
> 
>   * Fix environment variable on macOS to make dmg package work out of box
> 
> -- File Changes --
> 
>     M contrib/macos/FriCAS.c (4)
> 
> -- Patch Links --
> 
> https://github.com/fricas/fricas/pull/89.patch
> https://github.com/fricas/fricas/pull/89.diff

I can not really comment about this patch.  But AFAICS it affects only
dmg package.  So assuming it works for you just go on and commit.

-- 
                              Waldek Hebisch

PR #91, opened 2022-06-13
Fix build with CLISP-2.49.92

oldk1331:
Finally I have build FriCAS successfully with CLISP-2.49.92.

1. I can not build FriCAS with CLISP-2.49. (which is released in 2010).
2. I can build FriCAS with CLISP-2.49.92 (latest snapshot, released in 2018) or CLISP git version. (the development seems to be stalled, however.)
3. The CLISP must be built without threads support. If built with `--with-threads=POSIX_THREADS`, errors will happen when compiling SPAD files.

Some notes about the patch:
1. `ext:make-dir` is deprecated in 2007 and removed in 2017. So use `ext:make-directory` instead.
2. `interpsys` can not wirte to `$algebraOutputStream` because CLISP think it is closed. So initialize it again.

Please give this patch a review.  And if possible, test it on your system and see if CLISP can build FriCAS on your system.

I will update `INSTALL` section on CLISP later.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -476,6 +476,8 @@ Known problems
   Alternatively for sbcl one can use "--dynamic-space-size" argument
   to decrease use of virtual memory.
 
+- CLISP built with threads support may fail to compile FriCAS.
+
 - On new Linux kernel build using Clisp may take very long time.  This
   is caused by frequent calls to "fsync" performed without need by
   Clisp.
--- a/src/doc/sphinx/source/install.rst
+++ b/src/doc/sphinx/source/install.rst
@@ -524,6 +524,8 @@ Known problems
   Alternatively for sbcl one can use ``--dynamic-space-size`` argument
   to decrease use of virtual memory.
 
+- CLISP built with threads support may fail to compile FriCAS.
+
 - On new Linux kernel build using Clisp may take very long time.  This
   is caused by frequent calls to ``fsync`` performed without need by
   Clisp.
--- a/src/interp/util.lisp
+++ b/src/interp/util.lisp
@@ -326,10 +326,18 @@ After this function is called the image is clean and can be saved.
 (DEFVAR CUROUTSTREAM *standard-output*)
 
 (defun fricas-restart ()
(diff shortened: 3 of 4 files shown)


oldk1331:
Updated, please review.

BTW, I wonder if building in RAM disk (tmpfs) will avoid the 'fsync' slowness.

oldk1331:
Fails on git version of CLISP.  So it's better and simpler to drop support for `MAKE-DIR` at all.

````
*** - READ from
       #<INPUT BUFFERED FILE-STREAM CHARACTER
         #P"/home/runner/work/fricas/fricas/src/lisp/fricas-lisp.lisp" @884>
      : #<PACKAGE EXT> has no external symbol with name "MAKE-DIR"
The following restarts are available:
ABORT          :R1      Abort main loop
````

hebisch:
On Sat, Jun 18, 2022 at 10:43:20PM -0700, Qian Yun wrote:
> Fails on git version of CLISP.  So it's better and simpler to drop support for `MAKE-DIR` at all.
> 
> ````
> *** - READ from
>        #<INPUT BUFFERED FILE-STREAM CHARACTER
>          #P"/home/runner/work/fricas/fricas/src/lisp/fricas-lisp.lisp" @884>
>       : #<PACKAGE EXT> has no external symbol with name "MAKE-DIR"
> The following restarts are available:
> ABORT          :R1      Abort main loop
> ````

Well, I wrote the following in a message that apparently disappeard
somehere inside github:

AFAICS merely writing 'ext:foo-bar causes error in clisp.  So we
need to be more careful, like (untested, but similar to what we
use for different function in ECL):

     #+:clisp
     (defun makedir (fname)
            (let ((sym (or (find-symbol "MAKE-DIRECTORY" "EXT")
                           (find-symbol "MAKE-DIR" "EXT"))))
                 (funcall sym (pad-directory-name (namestring fname)))))

Note: This takes first existing symbol and calls it.  If symbol
does not exist or is not callable it will fail (as it should).
In principle it would be safer to check if symbol is callable,
but the above should be good enough.

-- 
                              Waldek Hebisch

Issue #78, opened 2022-01-08
Integration causes memory leak of all memory


ValZapod:
Did not check master, but on lastest debian testing (1.3.7) this causes crazy memory leak that stays until the )quit happens.

integrate((1 + x^2)^2/((1 - x^2)*(1 - 6*x^2 + x^4)^(3/4)), x)

hebisch:
On Sat, Jan 08, 2022 at 02:29:39PM -0800, Valerii Zapodovnikov wrote:
> Did not check master, but on lastest debian testing (1.3.7) this causes crazy memory leak that stays until the )quit happens.
> 
> integrate((1 + x^2)^2/((1 - x^2)*(1 - 6*x^2 + x^4)^(3/4)), x)

I do not think this is memory leak.  Simply FriCAS can take
quite a lot of memory (and time) to do such integrals.
For this integral one can significantly decrease resource
use by doing it in followning way:

(1) -> setSimplifyDenomsFlag(true)

   (1)  false
                                                                Type: Boolean
(2) -> integrate((1 + x^2)^2/((1 - x^2)*(1 - 6*x^2 + x^4)^(3/4)), x)
 
   >> Error detected within library code:
   integrate: implementation incomplete (residue poly has multiple non-linear factors)

The "implementation incomplete" message means that integral eventually
hit unimplemented part of Risch algorithm (actually this part is due
to Trager).  There is work going on to improve situation but
currently FriCAS can not do this integral (there is reasonable
chance that it is non-elementary).

Concerning "crazy memory leak that stays until the )quit happens",
you probably saw SBCL message about running out of memory.
Unfortunatly, this is "normal" behaviour of SBCL and FriCAS
has no control there: SBCL typically detects lack of memory
during garbage collection and signals fatal error.  One may
resonably expect that garbage collector should do its work
without allocating any memory, but unfortunately this in
not the case with SBCL garbage collector.  Even more
unfortunately SBCL has quite low memory limit (1GB), so
if you intend to do larger calculations you should
increase memory limit (see FriCAS INSTALL file for more
info).  In this case it is not clear if bigger memory
limit would help (I am trying your example with 16GB
as memory limit, currently it uses 4.749g, but I do
not know if 16GB is enough).

Anyway:
- FriCAS may use a lot of memory, we are working on
  increasing efficiency.  'setSimplifyDenomsFlag(true)'
  may help (or may make things worse...).
- behaviour when memory is exhausted is really due
  to Lisp implementation, we have no control there.

[...]

PR #92, opened 2022-06-15
integ.input: remove Zolotarev integral, fixed typo

ValZapod:
See e.g. https://en.wikipedia.org/wiki/Yegor_Ivanovich_Zolotarev

There is really no need to test Zolotarev integral, this is a very common case, which is the first mentioned on https://en.wikipedia.org/wiki/Risch_algorithm

### DIFF
--- a/src/input/integ.input
+++ b/src/input/integ.input
@@ -61,9 +61,7 @@ testIntegrate("(1+x^(1/4))^(1/3)/(x^(1/2))", "x", "alg")
 testIntegrate("((-x-1)*log((x^2+x))^2+2*log(x))/(x+1)", "x", "elem")
 testIntegrate("x/sqrt((1-x^2)*(1-k^2*x^2))", "x", "alg")
 testIntegrate("sqrt(1/x+1)*sqrt(x)-sqrt(x+1)", "x", "alg")
--- Abel example
-testIntegrate("(5*x-1)/sqrt(x^4 + 2*x^2 - 4*x + 1)", "x", "alg")
--- Massetr and Zanier
+-- Masser and Zannier
 testIntegrate("5*x^2/sqrt(x^6 +x)", "x", "alg")
 
 -- used to fail due to wrong normalization


oldk1331:
Why should the integral be removed, even if it is a common case? The test is put there to detect regression.

You can update your commit and the typo fix can be accepted.

ValZapod:
Because this is the most common example, that is tested in many other
examples implicitly, and I am afraid many times. Also, it was added because
of my off the hand comments but I was misunderstood, this "Abel" (actually
Abel did not even come close to solving this in common case) result is not
really important. See:
https://groups.google.com/g/sci.math.symbolic/c/PWoKqemXtHo/m/JQ8nFt8gBAAJ

чт, 16 июн. 2022 г., 13:49 Qian Yun ***@***.***>:

oldk1331:
Hi, I just saw your posts on sci.math.symbolic.

To compute `integrate(x/(x^2 - 1/5 - 2*%i/5)/sqrt(x^3 - x), x)`, you need to do the following and it returns in 0.5 second on fricas-1.3.7:

````
setSimplifyDenomsFlag true
integrate(x/((x^2 - ((1 + 2*sqrt(-1))/5))*sqrt(x^3 - x)), x)
````

Note that `setSimplifyDenomsFlag` can be helpful for some algebraic integrations sometimes, and `%i` is not the same as `sqrt(-1)`:  `%i` causes integration in Domain `Expression(Complex(Integer))` while `sqrt(-1)` makes it stays in `Expression(Integer)`.

As for this patch, I'll leave it for Waldek to decide.

ValZapod:
> setSimplifyDenomsFlag true

I got caught again and with the same magical value not less!

https://github.com/fricas/fricas/issues/78#issuecomment-1008457879

hebisch:
On Wed, Jun 15, 2022 at 06:35:50AM -0700, Valerii Zapodovnikov wrote:
> See e.g. https://en.wikipedia.org/wiki/Yegor_Ivanovich_Zolotarev
> 
> 
> There is really no need to test Zolotarev integral, this is a very common case, which is the first mentioned on https://en.wikipedia.org/wiki/Risch_algorithm
> 
>   https://github.com/fricas/fricas/pull/92

AFAIK _this specific integral_ first appeared in work of Abel,
so calling it "Zolotarev integral" has not grounds.  Let me mention
that Chebyshev and Zolotarev works were significant improvement in
constructive direction on works Abel, but while basic properties
found by Abel are important for FriCAS algortithms, AFAIK work of
Zolotarev is not used (modern developement went in quite different
direction).

Unlike similar looking integrals this integral presents specific
difficulty.  We have quite small number of examples of such
difficulty and this one of simplest examples. So it makes
sense to have it in testsuite.

An extra general remark: some routines are relatively
strightforward and one can give few tests and hope
for full coverage.  Integration algorithm has a lot
of branches and special cases and it is hard to decide
if test is redundant (OK, one could try planting bugs
and see if/which tests fail).  And even if test is
redundant today, it may be useful in the future when
integrator is modified (hopefully improved).  So, sorry
but in my judgement having this test is is worth more
than the cost (2 lines in test file + excution time).

Thanks for spoting the typo, I will commit correction.
-- 
                              Waldek Hebisch

PR #95, opened 2022-06-19
load correct gmp library on windows

oldk1331:
Next question is whether we should bundle this gmp dll in our windows distribution.  I think the answer should be yes.


hebisch:
On Sun, Jun 19, 2022 at 07:38:45AM -0700, Qian Yun wrote:
> Next question is whether we should bundle this gmp dll in our windows distribution.  I think the answer should be yes.

Yes, please commit.
 
-- 
                              Waldek Hebisch

PR #96, opened 2022-06-19
use $spadroot instead of (|getEnv| "FRICAS")

oldk1331:
The design purpose is to store "$FRICAS" into `$spadroot`.  And when `$FRICAS` changes, run `reroot`.  So all these `(|getEnv| "FRICAS")` should be changed to use `$spadroot` or `make-absolute-filename`.

(Putting here for review. Merge can be waited after release.)

### DIFF
--- a/src/interp/br-data.boot
+++ b/src/interp/br-data.boot
@@ -197,13 +197,13 @@ dbHasExamplePage conname ==
   sname    := STRINGIMAGE conname
   abb      := constructor? conname
   ucname   := UPCASE STRINGIMAGE abb
-  pathname :=STRCONC(getEnv '"FRICAS",'"/share/hypertex/pages/",ucname,'".ht")
+  pathname :=STRCONC($SPADROOT,'"/share/hypertex/pages/",ucname,'".ht")
   isExistingFile pathname => INTERN STRCONC(sname,'"XmpPage")
   nil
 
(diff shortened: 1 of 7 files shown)


hebisch:
On Sun, Jun 19, 2022 at 08:05:17AM -0700, Qian Yun wrote:
> The design purpose is to store &quot;$FRICAS&quot; into `$spadroot`.  And when `$FRICAS` changes, run `reroot`.  So all these `(|getEnv| &quot;FRICAS&quot;)` should be changed to use `$spadroot` or `make-absolute-filename`.
> 
> (Putting here for review. Merge can be waited after release.)

I admit that it is not clear for me why you want this.  At least
some of places that you want to modify comunicate with C code.
And in C code all we have is FRICAS environment variable
(we definitely do not have access to Lisp '$spadroot').

IIUC you wrote that passing environment variables on Mac OS is
tricky.  Having some fallback is reasonable, but preferably
this fallback should be in one place, at top of process tree.
And once fallback is established it should be propagated
down via environment.  My feeling is that in longer term
is is better to learn how to propagate environment variables
on Mac OS.  Otherwise, we risk that hacks which are reasonable
as "point" solution spread out to many places creating
hard to maintain mess.

-- 
                              Waldek Hebisch

PR #98, opened 2022-06-23
use lisp function instead of shell commands for SBCL and ClozureCL

oldk1331:
This is useful when I try to do a pure lisp compile (aka don't use C functions, don't load shared library.)

Also this will works on windows as well, and performance is better.

### DIFF
--- a/src/interp/nlib.lisp
+++ b/src/interp/nlib.lisp
@@ -321,9 +321,9 @@
 
 #+:sbcl
 (defun delete-directory (dirname)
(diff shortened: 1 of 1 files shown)


hebisch:
On Wed, Jun 22, 2022 at 09:29:22PM -0700, Qian Yun wrote:
> This is useful when I try to do a pure lisp compile (aka don&#39;t use C functions, don&#39;t load shared library.)
> 
> Also this will works on windows as well, and performance is better.
> 
>   https://github.com/fricas/fricas/pull/98

OK, please commit.

-- 
                              Waldek Hebisch

PR #99, opened 2022-06-23
Use String instead of Symbol in BOOT code, especially for printing functions

oldk1331:
When I was exploring symbols in "BOOT" package, I found many strange symbols -- they are very long and contains space.  Turns out they should be strings instead -- they are not quoted in *.boot files.

So I fixed some of them.  If this is the correct approach, I can fix the remaining 200 or so.

### DIFF
--- a/src/interp/apply.boot
+++ b/src/interp/apply.boot
@@ -203,7 +203,7 @@ compMapCond(op, mc, bindings, fnsel, e) ==
 
 compMapCond'([cexpr,fnexpr], op, dc, bindings, e) ==
   compMapCond''(cexpr, dc, e) => compMapCondFun(fnexpr,op,dc,bindings)
-  stackMessage ["not known that",'%b,dc,'%d,"has",'%b,cexpr,'%d]
+  stackMessage ['"not known that",'%b,dc,'%d,'"has",'%b,cexpr,'%d]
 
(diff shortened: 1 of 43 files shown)


hebisch:
On Thu, Jun 23, 2022 at 04:06:25AM -0700, Qian Yun wrote:
> When I was exploring symbols in &quot;BOOT&quot; package, I found many strange symbols -- they are very long and contains space.  Turns out they should be strings instead -- they are not quoted in *.boot files.
> 
> So I fixed some of them.  If this is the correct approach, I can fix the remaining 200 or so.

Looks good.  Just little remark: it is not always easy to decide
which one should be strings and which should be symbols, but
the cases you changed are resonably clear.

-- 
                              Waldek Hebisch

oldk1331:
I have updated second batch.

I find "symbols with space in them" by `grep -e " '|[^|]* [^|]*|" *clisp`.

So this method is relatively safe, but of course, will miss some other cases.

When I was doing the edits, I notice the function to make sure it takes strings as argument.

After this commit, I reduce the number of symbols in BOOT package from 22587 to 22324.
(Found by `)lisp (length (loop for s being the symbols of (find-package "BOOT") collect s))`)

I'll leave this open for review for a few days, and merge this after that.

PR #100, opened 2022-06-26
disable GitHub linguist detection because it is inaccurate

oldk1331:
There was discussion in #53 and #58, the problem is that GitHub wrongly flags some source files to Clojure.

We didn't reach an agreement on how to flag our sources.

GitHub doesn't allow us to flag sources as we want.  And BOOT and SPAD are used only in FriCAS, so it is not qualified to put them into GitHub's language database either.

I think we can all agree on disable this inaccurate language detection statistics.  Have nothing is better than wrong and misleading.  We can add back proper support when things change.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
 * text=auto eol=lf
+* -linguist-detectable


hemmecke:
Switching it off altogether makes sense to me. +1

hebisch:
On Sat, Jun 25, 2022 at 06:35:36PM -0700, Qian Yun wrote:
> There was discussion in #53 and #58, the problem is that GitHub wrongly flags some source files to Clojure.
> 
> We didn&#39;t reach an agreement on how to flag our sources.
> 
> GitHub doesn&#39;t allow us to flag sources as we want.  And BOOT and SPAD are used only in FriCAS, so it is not qualified to put them into GitHub&#39;s language database either.
> 
> I think we can all agree on disable this inaccurate language detection statistics.  Have nothing is better than wrong and misleading.  We can add back proper support when things change.

OK.  Disabling wrong detection is the best we can do now (and
likely will remain best possibility in the future).

-- 
                              Waldek Hebisch

Issue #59, opened 2021-05-28
build on macOS with ecl fails due to strict C compiler


dimpase:
see https://trac.sagemath.org/ticket/31863
typical errors are
```
;;; End of Pass 1.
;;; Internal error:
;;;   ** Error code 1 when executing
;;; (EXT:RUN-PROGRAM "gcc" ("-I." "-I/Users/palmieri/Desktop/Sage/git/sage/local/include/" "-O2" "-g" "-march=native" "-fPIC" "-fno-common" "-Ddarwin" "-O2" "-c" "fricas-lisp.c" "-o" "fricas-lisp.o")):
;;; fricas-lisp.c:608:25: error: implicit declaration of function 'writeablep' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
;;;   value0 = ecl_make_int(writeablep(ecl_base_string_pointer_safe(v2)));
;;;                         ^
;;; /Users/palmieri/Desktop/Sage/git/sage/local/include/ecl/external.h:1107:23: note: expanded from macro 'ecl_make_int'
;;; # define ecl_make_int ecl_make_int32_t
;;;                       ^
;;; fricas-lisp.c:639:25: error: implicit declaration of function 'remove_directory' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
;;;   value0 = ecl_make_int(remove_directory(ecl_base_string_pointer_safe(v2)));
;;;                         ^
;;; /Users/palmieri/Desktop/Sage/git/sage/local/include/ecl/external.h:1107:23: note: expanded from macro 'ecl_make_int'
;;; # define ecl_make_int ecl_make_int32_t
;;;                       ^
;;; fricas-lisp.c:639:25: note: did you mean 'L25_remove_directory_'?
;;; /Users/palmieri/Desktop/Sage/git/sage/local/include/ecl/external.h:1107:23: note: expanded from macro 'ecl_make_int'
;;; # define ecl_make_int ecl_make_int32_t
;;;                       ^
;;; ./fricas-lisp.eclh:35:18: note: 'L25_remove_directory_' declared here
;;; static cl_object L25_remove_directory_(cl_object );
;;;                  ^
;;; fricas-lisp.c:670:25: error: implicit declaration of function 'open_server' is invalid in C99 [-Werror,-Wimplicit-function-declaration]
;;;   value0 = ecl_make_int(open_server(ecl_base_string_pointer_safe(v2)));
;;;                         ^
[...]

PR #102, opened 2022-06-28
GH Actions workflow for continuous integration and portability testing

mkoeppe:
Motivated by the SageMath / passagemath distributions, which include FriCAS as an optional package (using ECL and targeting Linux and macOS), this PR adds a continuous integration workflow.

Test runs: https://github.com/passagemath/fricas/actions/workflows/ci-sage.yml

### DIFF
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -81,7 +81,7 @@ jobs:
       - name: Run make dist, prepare upstream artifact
         # see top-level Makefile.in: Need a full out-of-source build before running 'make dist'
         run: |
-          (cd build/pkgs/${{ env.SPKG }}/src && mkdir build && cd build && ../configure --srcdir=.. --with-lisp=sbcl --with-gmp && make -j2 --output-sync && make dist) \
+          (cd build/pkgs/${{ env.SPKG }}/src && mkdir build && cd build && ../configure --srcdir=.. --with-lisp=sbcl --with-gmp && make && make dist) \
           && mkdir -p upstream && cp build/pkgs/${{ env.SPKG }}/src/build/*.tar.bz2 upstream/${{ env.SPKG }}-git.tar.bz2 \
           && echo "sage-package create ${{ env.SPKG }} --version git --tarball ${{ env.SPKG }}-git.tar.gz --type=optional" > upstream/update-pkgs.sh \
           && if [ -n "${{ env.REMOVE_PATCHES }}" ]; then echo "(cd ../build/pkgs/${{ env.SPKG }}/patches && rm -f ${{ env.REMOVE_PATCHES }}; :)" >> upstream/update-pkgs.sh; fi \
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -79,9 +79,10 @@ jobs:
           sudo DEBIAN_FRONTEND=noninteractive apt-get update
           sudo DEBIAN_FRONTEND=noninteractive apt-get install $DIST_PREREQ
       - name: Run make dist, prepare upstream artifact
+        # see top-level Makefile.in: Need a full out-of-source build before running 'make dist'
         run: |
-          (cd build/pkgs/${{ env.SPKG }}/src && ./configure --enable-case-insensitive-file-system-check=no && make dist) \
-          && mkdir -p upstream && cp build/pkgs/${{ env.SPKG }}/src/*.tar.gz upstream/${{ env.SPKG }}-git.tar.gz \
+          (cd build/pkgs/${{ env.SPKG }}/src && mkdir build && cd build && ../configure --srcdir=.. --with-lisp=sbcl --with-gmp && make && make dist) \
+          && mkdir -p upstream && cp build/pkgs/${{ env.SPKG }}/src/build/*.tar.gz upstream/${{ env.SPKG }}-git.tar.gz \
           && echo "sage-package create ${{ env.SPKG }} --version git --tarball ${{ env.SPKG }}-git.tar.gz --type=optional" > upstream/update-pkgs.sh \
           && if [ -n "${{ env.REMOVE_PATCHES }}" ]; then echo "(cd ../build/pkgs/${{ env.SPKG }}/patches && rm -f ${{ env.REMOVE_PATCHES }}; :)" >> upstream/update-pkgs.sh; fi \
           && ls -l upstream/
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -79,11 +79,10 @@ jobs:
           sudo DEBIAN_FRONTEND=noninteractive apt-get update
           sudo DEBIAN_FRONTEND=noninteractive apt-get install $DIST_PREREQ
       - name: Run make dist, prepare upstream artifact
-        # see top-level Makefile.in: Need a full out-of-source build before running 'make dist'
         run: |
-          (cd build/pkgs/${{ env.SPKG }}/src && mkdir build && cd build && ../configure --srcdir=.. --with-lisp=sbcl --with-gmp && make && make dist helpsrcdir=src/doc) \
-          && mkdir -p upstream && cp build/pkgs/${{ env.SPKG }}/src/build/*.tar.bz2 upstream/${{ env.SPKG }}-git.tar.bz2 \
-          && echo "sage-package create ${{ env.SPKG }} --version git --tarball ${{ env.SPKG }}-git.tar.gz --type=optional" > upstream/update-pkgs.sh \
+          (cd build/pkgs/${{ env.SPKG }}/src && git archive --format=tar --prefix=${{ env.SPKG }}-git/ HEAD | bzip2 > ${{ env.SPKG }}-git.tar.bz2 ) \
+          && mkdir -p upstream && cp build/pkgs/${{ env.SPKG }}/src/*.tar.bz2 upstream/${{ env.SPKG }}-git.tar.bz2 \
+          && echo "sage-package create ${{ env.SPKG }} --version git --tarball ${{ env.SPKG }}-git.tar.bz2 --type=optional" > upstream/update-pkgs.sh \
           && if [ -n "${{ env.REMOVE_PATCHES }}" ]; then echo "(cd ../build/pkgs/${{ env.SPKG }}/patches && rm -f ${{ env.REMOVE_PATCHES }}; :)" >> upstream/update-pkgs.sh; fi \
           && ls -l upstream/
       - uses: actions/upload-artifact@v2
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -80,7 +80,7 @@ jobs:
           sudo DEBIAN_FRONTEND=noninteractive apt-get install $DIST_PREREQ
       - name: Run make dist, prepare upstream artifact
         run: |
-          (cd build/pkgs/${{ env.SPKG }}/src && ./configure && make dist) \
+          (cd build/pkgs/${{ env.SPKG }}/src && ./configure --enable-case-insensitive-file-system-check=no && make dist) \
           && mkdir -p upstream && cp build/pkgs/${{ env.SPKG }}/src/*.tar.gz upstream/${{ env.SPKG }}-git.tar.gz \
           && echo "sage-package create ${{ env.SPKG }} --version git --tarball ${{ env.SPKG }}-git.tar.gz --type=optional" > upstream/update-pkgs.sh \
           && if [ -n "${{ env.REMOVE_PATCHES }}" ]; then echo "(cd ../build/pkgs/${{ env.SPKG }}/patches && rm -f ${{ env.REMOVE_PATCHES }}; :)" >> upstream/update-pkgs.sh; fi \
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -56,7 +56,7 @@ env:
   # Sage distribution packages to build
   TARGETS_PRE:      build/make/Makefile
   TARGETS:          SAGE_CHECK=no SAGE_CHECK_PACKAGES=fricas fricas
-  TARGETS_OPTIONAL:
+  TARGETS_OPTIONAL: build/make/Makefile
   # Standard setting: Test the current beta release of Sage:
   SAGE_REPO:        sagemath/sage
   SAGE_REF:         develop
--- a/.github/workflows/ci-sage.yml
+++ b/.github/workflows/ci-sage.yml
@@ -1,0 +1,259 @@
+name: Run Sage CI for Linux/Cygwin/macOS
+
+## This GitHub Actions workflow provides:
+##
+##  - portability testing, by building and testing this project on many platforms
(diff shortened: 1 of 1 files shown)


mkoeppe:
Currently `make dist` is failing:
```
2022-06-28T19:43:52.4519851Z test -f /Integer.help \
2022-06-28T19:43:52.4520225Z || (echo "helpfiles are missing"; \
2022-06-28T19:43:52.4520899Z     echo "use 'make dist helpsrcdir=/path/to/spadhelp'"; \
2022-06-28T19:43:52.4521241Z     exit 1)
2022-06-28T19:43:52.4536747Z helpfiles are missing
2022-06-28T19:43:52.4537345Z use 'make dist helpsrcdir=/path/to/spadhelp'
2022-06-28T19:43:52.4541354Z make: *** [Makefile:417: help-sanity-check] Error 1
2022-06-28T19:43:52.4562986Z ##[error]Process completed with exit code 2.
```
Any hints how to fix this?

mkoeppe:
I also ran the build locally (on macOS), and no `*.help` files are generated in any directory. Is `make dist` actually being used for making `fricas` source distributions?

hebisch:
On Tue, Jun 28, 2022 at 01:54:02PM -0700, Matthias K??ppe wrote:
> I also ran the build locally (on macOS), and no `*.help` files are generated in any directory. Is `make dist` actually being used for making `fricas` source distributions?

Help files duplicate data from other documentation files.  In principle
they could be made automatically from other files, but nobody
implemented this.  Current help files were created by hand.
They are not in repository as the intent is either have automatic
process or drop them if they get out of date.

And yes, `make dist` is used, it just takes help files from
previous release...

-- 
                              Waldek Hebisch

hebisch:
On Tue, Jun 28, 2022 at 01:54:02PM -0700, Matthias K??ppe wrote:
> I also ran the build locally (on macOS), and no `*.help` files are generated in any directory. Is `make dist` actually being used for making `fricas` source distributions?

Sorry, for confusion in previous answer.  I make releases using
'src/scripts/mkdist.sh'.  I do not use 'make dist'.

-- 
                              Waldek Hebisch

hemmecke:
> Help files duplicate data from other documentation files. In principle they could be made automatically from other files, but nobody implemented this.

What about 
https://github.com/fricas/fricas/blob/master/src/doc/Makefile.in#L528
? I am sure I automated this.
If "make  stamp-ug16helpfiles" does not work, you can simply run the awk script by hand.
```
awk -f src/doc/syscmd.awk src/doc/htex/ug16.htex 
```

mkoeppe:
I ran `make stamp-ug16helpfiles` in `build/src/doc` manually, and it did generate some `*.help` files, but it did not generate `Integer.help`, which `make help-sanity-check` is looking for

mkoeppe:
> I make releases using 'src/scripts/mkdist.sh'.

How does one use `src/scripts/mkdist.sh`?

mkoeppe:
In particular, `src/scripts/mkdist.sh` does not work in an out-of-tree configured source tree:
```
cp: /Users/mkoeppe/s/sage/fricas/build/../build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist/build/dist: File name too long
```

mkoeppe:
Also does not work in an in-tree configured source tree:
{{{
fricas ➤ src/scripts/mkdist.sh                                                                                                                                                              git:ci_sage
SRCDIR=/Users/mkoeppe/s/sage/fricas
copy_lisp=""
copy_gphts=""
copy_phts=""
GCL_DIST=""
HELP_DIR=""
cp: /Users/mkoeppe/s/sage/fricas/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/dist/license/LICENSE.AXIOM: File name too long
}}}

hebisch:
On Tue, Jun 28, 2022 at 04:49:14PM -0700, Matthias K??ppe wrote:
> In particular, `src/scripts/mkdist.sh` does not work in an out-of-tree configured source tree:

It is not clear to me what you did.  Normal usage is as follows:

1) build out-of-source-tree
2) _in the build directory_ run

/path/to/source/tree/src/scripts/mkdist.sh --copy_lisp --copy_phts --copy_help=/path_to_dir_containing_help_files


As result you will get subdirectory called 'dist'.  This subdirectory
is than renamed and packed using tar to get distribution tarball
(both steps done by hand).

If you skip --copy_lisp tarball will contain no precompiled Lisp.
If you skip --copy_phts there will be no pre-generated HyperDoc
pages.  If you skip --copy_help there will be no help files
in the tarball.

Note: 'mkdist.sh' uses recursive copy, it will fail if you try
to run it inside source tree.

-- 
                              Waldek Hebisch

mkoeppe:
I did 1 & 2 but then called `src/scripts/mkdist.sh` without arguments. I'll try again with these arguments

mkoeppe:
Given that `make dist` is no longer used and appears to be broken, would you accept a pull request that replaces it by a clear error message pointing to the `mkdist.sh` script?

hebisch:
On Tue, Jun 28, 2022 at 02:57:24PM -0700, Ralf Hemmecke wrote:
> > Help files duplicate data from other documentation files. In principle they could be made automatically from other files, but nobody implemented this.
> 
> What about 
> https://github.com/fricas/fricas/blob/master/src/doc/Makefile.in#L528
> ? I am sure I automated this.
> If "make  stamp-ug16helpfiles" does not work, you can simply run the awk script by hand.
> ```
> awk -f src/doc/syscmd.awk src/doc/htex/ug16.htex 
> ```

mkdist.sh picks generated help files.  But there are several
more which are _not_ generated.

-- 
                              Waldek Hebisch

oldk1331:
Can you tell me the purpose of this PR?  Looks like it builds on 135 different systems.  Could take hours if not days to finish.

On the other hand, FriCAS has very few dependencies and extremely portable among different distros.

Also this only tests ECL.  I think the added value is minor compared to the effort.

mkoeppe:
There's no need to worry about the effort that I'm putting in it.

oldk1331:
The effort I meant is in Eco-friendly sense.  So this is 135x40minutes tests compared with current 3x10minutes.  Compared with 180 times more time/energy, the gained value seems small.

Also, would this fits more in sage repo?

mkoeppe:
> would this fit more in sage repo?

We already run this portability suite in the Sage repo. However, when we prepare a package upgrade, such as fricas to 1.3.8 in https://trac.sagemath.org/ticket/34051, we often discover that packages break one or several platforms. The purpose of contributing `ci-sage.yml` to an upstream project is to assist upstream projects in discovering portability breakage before they cut a new release.

Our `ci-sage.yml` is already in use with various mathematical software (see https://trac.sagemath.org/ticket/33338) for this purpose.

mkoeppe:
> FriCAS has very few dependencies and extremely portable among different distros.

Well, Sage has carried patches for FriCAS because it does not compile out of the box on all of our supported platforms. https://github.com/sagemath/sage/tree/develop/build/pkgs/fricas/patches

And it appears that FriCAS currently does not have any CI coverage for builds with ECL. As they say, untested, hence broken

mkoeppe:
> The effort I meant is in Eco-friendly sense. So this is 135x40minutes tests

It's easy to configure it so it only runs on tags, etc., if so desired; or to reduce the number of platform tested.

oldk1331:
> We already run this portability suite in the Sage repo. However, when we prepare a package upgrade, such as fricas to 1.3.8 in https://trac.sagemath.org/ticket/34051, we often discover that packages break one or several platforms.

So for 1.3.8, specifically, what are the problems?

From the ticket#34051, see the problems happen if you remove the patch?

Well, if you remove https://github.com/sagemath/sage/blob/develop/build/pkgs/fricas/patches/remove_case_insensitive_test.patch , it is fine on released tarball, but will fail on git version.

About macOS failure, I wonder if it is the same as https://github.com/fricas/fricas/issues/59 , fixed in ECL upstream but not released?

mkoeppe:
Ticket https://trac.sagemath.org/ticket/34051 removes some of the patches. https://trac.sagemath.org/ticket/34051#comment:14 proposes to add a reduced version of one of the patches back.

You can see the macOS failure for example in https://github.com/mkoeppe/fricas/runs/7104014683?check_suite_focus=true

My hope would be that versions of our patches can be upstreamed so that we don't have to carry and update them any more.
If these portability fixes are something that FriCAS developers are interested in, then the CI in this PR may be a valuable tool.

oldk1331:
> You can see the macOS failure for example in https://github.com/mkoeppe/fricas/runs/7104014683?check_suite_focus=true

Where can I get the FriCAS build log in this link?

mkoeppe:
Open the section "Print out logs for immediate inspection" in the log output

oldk1331:
OK, confirmed, same problem as https://github.com/fricas/fricas/issues/59 .

````
Re implicit declarations there are two ways around this problem:

    1. use dynamic ffi (a bonus point is that it will work when ecl is compiled only with the bytecodes compiler, downside is a big performance penalty - around 40x slower function call)
    2. expand your macro to add a declaration (that's what cffi does for static calls) or include the appropriate header
````

We didn't move forward after this suggestion from ECL developer.

oldk1331:
New error from Fedora-34,35,36,37 standard:

````
2022-06-29T08:47:01.7064176Z ;;; Internal error:
2022-06-29T08:47:01.7064263Z ;;;   ** Error code 1 when executing
2022-06-29T08:47:01.7065551Z ;;; (EXT:RUN-PROGRAM "gcc" ("-o" "ARR2CAT-.NRLIB/ARR2CAT-.fas" "-L/usr/lib64/" "/sage/local/var/tmp/sage/build/fricas-git/src/src/algebra/ARR2CAT-.NRLIB/ARR2CAT-.o" "-shared" "-Wl,-z,relro" "-Wl,--as-needed" "-Wl,-z,now" "-specs=/usr/lib/rpm/redhat/redhat-hardened-ld" "-specs=/usr/lib/rpm/redhat/redhat-annobin-cc1" "-Wl,--build-id=sha1" "-Wl,-z,relro" "-Wl,--as-needed" "-Wl,-z,now" "-specs=/usr/lib/rpm/redhat/redhat-hardened-ld" "-specs=/usr/lib/rpm/redhat/redhat-annobin-cc1" "-Wl,--build-id=sha1" "-lecl" "-lgmp" "-lgc" "-lffi" "-lgc" "-lgc" "-lgc" "-lpthread" "-ldl" "-lm")):
2022-06-29T08:47:01.7065702Z ;;; make[8]: *** write jobserver: Bad file descriptor.  Stop.
2022-06-29T08:47:01.7065828Z ;;; make[8]: *** Waiting for unfinished jobs....
2022-06-29T08:47:01.7065970Z ;;; make[8]: *** write jobserver: Bad file descriptor.  Stop.
2022-06-29T08:47:01.7066394Z ;;; lto-wrapper: fatal error: make returned 2 exit status
2022-06-29T08:47:01.7066497Z ;;; compilation terminated.
2022-06-29T08:47:01.7066681Z ;;; /usr/bin/ld: error: lto-wrapper failed
2022-06-29T08:47:01.7066805Z ;;; collect2: error: ld returned 1 exit status
````

Seems like lto link error?

Also this failure to compile should lead to later interpsys crash: `Detected access to an invalid or protected memory address.` `No such file or directory`

mkoeppe:
> More generally about CI: CI is good at detecting that there are problems, but IME of limited help in fixing them.

Yes, developers still got to develop.

mkoeppe:
> New error from Fedora-34,35,36,37 standard:

By the way, to investigate these failures more, you can pull the failing builds as Docker images.
For example `docker run -it docker.pkg.github.com/mkoeppe/fricas/sage-docker-fedora-36-standard-with-targets:4e12f4003a-dirty-failed bash`

ValZapod:
>The effort I meant is in Eco-friendly sense.

Fricas is nothing compared to FFmpeg, Chromium/Android, and Wireshark or Rawtherapee. All of them build new versions every 24 hours or for every commit, including pull requests. Cause all of them are very hard to compile and cause it may be impossible to compile withlut installing the older compiler and older glibc, etc. Or just to test pull requests.

Literally: https://github.com/ValZapod/FFmpeg-Builds/tree/master

https://www.wireshark.org/download/automated/win64/

https://chromium-review.googlesource.com/q/status:open+-is:wip

https://android-review.googlesource.com/q/status:open+-is:wip

https://github.com/Beep6581/RawTherapee/actions

hebisch:
On Sat, Mar 25, 2023 at 04:10:14PM -0700, Dima Pasechnik wrote:
> > So for 1.3.8, specifically, what are the problems?
> 
> for instance, it does not work on ECL macOS without our patches. Which were offered here, by the way, I don't know why you cannot get around to merge them.

Sorry for late replay.  _Parts_ of your patches are commited.  Concerning
declarations for ECL FriCAS in git repo generates extern declarations
for C functions.  This is less work for developers and should be more
robust than your proposal with includes.  The extern part was commited
after 1.3.8, so if you test only release you will see it in next release,
but other folks reported that the code works as intended on macOS.

-- 
                              Waldek Hebisch

PR #111, opened 2022-10-28
Introduced a way to hardcode SBCL_HOME

catap:
I've added a configure argument `--with-sbcl-home` which allows to specify desired `SBCL_HOME`.

Without it fricas' lisp can't be really used without specification of `SBCL_HOME` environment variable at some cases that leads to wired issue like:
```
(1) -> )lisp (require :asdf)

   >> System error:
   Don't know how to REQUIRE ASDF.
See also:
  The SBCL Manual, Variable SB-EXT:*MODULE-PROVIDER-FUNCTIONS*
  The SBCL Manual, Function REQUIRE

(1) ->
```

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -142,6 +142,9 @@ install-src:
 	echo FRICAS_VERSION='"${PACKAGE_VERSION}"' >> '${COMMAND}'.tmp
 	echo FRICAS_LISP_FLAVOR='"${FRICAS_LISP_FLAVOR}"' >> '${COMMAND}'.tmp
 	echo FRICAS_LISP_VERSION='"${FRICAS_LISP_VERSION}"' >> '${COMMAND}'.tmp
+	if test "x$(SBCL_HOME)" != "x" ; then \
+		echo export SBCL_HOME='"$${SBCL_HOME:-$(SBCL_HOME)}"' >> '${COMMAND}'.tmp ; \
+	fi
 	cat $(fricas_src_srcdir)/etc/fricas >> '${COMMAND}'.tmp
 	$(INSTALL_SCRIPT) '${COMMAND}'.tmp '${COMMAND}'
 	-rm '${COMMAND}'.tmp
--- a/config/var-def.mk
+++ b/config/var-def.mk
@@ -170,6 +170,8 @@ FASLEXT = @fricas_fasl_type2@
 FRICAS_LISP_FLAVOR=@fricas_lisp_flavor@
 FRICAS_LISP_VERSION=@fricas_lisp_version@
 
+SBCL_HOME=@sbcl_home@
+
 ##
 
 ##
--- a/configure
+++ b/configure
@@ -663,6 +663,7 @@ GMP_LIBDIR
 GMP_LDFLAGS
 GMP_CPPFLAGS
 GMP_WRAP_SO_TARGET
+sbcl_home
 SOLIB_FLAGS
 LIBSPAD_SO_TARGET
 fricas_lisp_version
@@ -752,6 +753,7 @@ enable_case_insensitive_file_system_check
 with_pre_generated
 with_lisp
 with_lisp_flavor
+with_sbcl_home
 with_gmp
 with_gmp_include
 with_gmp_lib
@@ -1419,6 +1421,7 @@ Optional Packages:
                           code.
   --with-lisp=L           use L as Lisp platform
   --with-lisp-flavor=F    obsolete and ignored
+  --with-sbcl-home=PATH   specify the path of hardcoded SBCL_HOME
   --with-gmp=PATH         specify prefix directory for the installed GMP
                           package. Equivalent to
                           --with-gmp-include=PATH/include plus
@@ -4199,6 +4202,17 @@ case $target in
        SOLIB_FLAGS="-shared"
          ;;
 esac
+
+
+
+## Configure desired SBCL_HOME
+
+# Check whether --with-sbcl-home was given.
+if test "${with_sbcl_home+set}" = set; then :
+  withval=$with_sbcl_home; sbcl_home=$withval
+else
+  sbcl_home=""
+fi
 
 
 
--- a/configure.ac
+++ b/configure.ac
@@ -325,6 +325,13 @@ esac
 AC_SUBST(LIBSPAD_SO_TARGET)
 AC_SUBST(SOLIB_FLAGS)
 
+## Configure desired SBCL_HOME
+AC_ARG_WITH([sbcl-home],
+    AS_HELP_STRING([--with-sbcl-home=PATH], [specify the path of hardcoded SBCL_HOME]),
+    sbcl_home=$withval,
+    sbcl_home="")
+AC_SUBST(sbcl_home)
+
 ## In case our underlying lisp is SBCL or Closure CL, we provide a
 ## wrapper to enable GMP bignums in lisp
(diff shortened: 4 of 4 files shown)


hemmecke:
As far as I remember, FRICASsys lost the default information about SBCL_HOME (which the original sbcl image has). So it might make sense to support setting SBCL_HOME.

However, as Qian said, there are ideas to support asdf in fricas.

To my taste, if the setting of SBCL_HOME should be in configure.ac, then it should result in an error if the underlying lisp flavour is not sbcl. Furthermore, the installation directory is not (very) hard-coded in the (binary) distribution. Which means, that one can easily move the it to some other directory, change the directories in the shell scripts, fricas and efricas and the relocated distribution would still work. It is questionable what to do with SBCL_HOME in this case, since sbcl on the client side might live at a different place as on the developer side, i.e., setting SBCL_HOME makes only sense for someone who compiles fricas on his own computer.

hemmecke:
BTW, I think it should work if you add "export SBCL_HOME=..." to your fricas script.

catap:
@hemmecke / @oldk1331 I'm using sbcl which is installed via MacPorts, and I'm running fricas from jupyter via jfricas.

I have no idea how can I add `SBCL_HOME` to jupyter's kernel. Ok, I may hack / add a patch to jfricas, but it seems quite wired because fricas might be build with another lisp.

Anyway, jfricas calls fricas, and it seems the simple solution just add the right SBCL_HOME when it required.

MacPorts installs everything into /opt/local that makes things more tricky, and it supports different installation prefix.

I may keep this patch as part of MacPorts' port, but I preferer to send it to upstream.

If you ok with it I'll update it to fail when used lisp isn't sbcl.

hemmecke:
I don't understand, what this has to do with jfricas. Unfortunately, the current jfricas does not always work, because of finding hunchentoot and if this is not in the right place. Maybe this is the reason why you asked for SBCL_HOME. Is it?
If not, then why do you need asdf?
And no, I don't think that updating the patch would be useful.

We rather want to produce a distribution that already includes hunchentoot in the FRICASsys image, so loading the webserver via asdf would be unnecessary. If you are interested in that then maybe you motivate me a little.

catap:
@hemmecke sorry for very long response. I've missed your comment :(

Right now I'm using this patch to load webserver which is instabled by MacPorts inside FRICAS to enjoy FRICAS and Jupyter integration.

I not sure that build a bundel is the best approach for MacPorts like instalation.

hemmecke:
You have probably realized that FriCAS 1.3.9 is out. That includes a few routines that were first in jfricas and moved to the official repo because this is where they belong.
Second, forget about my previous approach and SBCL_HOME and such. Not the installation relies on the fact the FriCAS was built with Hunchentoot included. You can find a script to build such hsbcl from https://github.com/fricas/fricas/releases/tag/1.3.9 
Follow the steps at http://fricas.github.io/install.html to install fricas and jfricas (actually now you can install jfricas even via "pip install jfricas").
Maybe you subscribe to our mailing list https://groups.google.com/g/fricas-devel, because most communication about fricas happens there.

catap:
I see.

So, you're shipping some version of hunchentoot and all its dependencies as separated bundle and just creates some wrapper to be used.

I not sure that it is good way to install it, nor it is accepted by any major distributive like Debian.

From my point of view using  `SBCL_HOME` to allow using system's hunchentoot is the best and simpler way.

Why? At least from security point of view. Let assume that some very bad security issue had been discovered  inside hunchentoot or bordeaux-threads or any dependecies.

If you're using system's library the time to update is much shortest. Why? Because the maintainer of compoment will be notified about security issues, and not everyone who builds a bundle with that component.

So, my question remains: how is it possible to use system installed libraries for jfricas?

catap:
Anyway, it is clear that this PR is no go => let me close it.

hemmecke:
Oh, you have probably not looked into https://github.com/fricas/fricas/releases/download/1.3.9/hsbcl-1.3.9.tar . Yes, it comes with a collection of zip files. to build hunchentoot. In fact, you should see that tar file as a way of producing a working version of an sbcl image with hunchentoot included. Nobody prevents you from using other libraries of system provided libraries to build hsbcl. You just have to rebuild FriCAS from source on top of this (which would not be much different from the way I described initially, before Waldek came up with the hsbcl idea.
The only problem may be that the fricas binary is a frozen distribution (including hunchentoot in the linux version). That will not change if there were issues with hunchentoot. But that's life with lisp images.

catap:
@hemmecke lisp images is like static linking of binary. I understand why it's happened but I'm pointing that it isn't good idea for complex system with a lot of dependencies.

FriCAS itself as far as I know hasn't got any lisp dependencies => here I may rationalise usage of kind of static linking. It doesn't change anything.

Well, it need to be rebuild after change of sbcl but it happens rarely.

Anyway, adding hunchentooth which has dozen dependencies makes things stranger.

From my point of view exposing `SBCL_HOME` and/or `asdf` is the right way to support such dependencies. Quite universal and for any lisp which supports `asdf` and FriCAS.

But I agree that hunchentooth the first and probably the last dependencies of that kind.

To be clear: I've upgraded everything inside MacPorts, and removed my old patch to follow an approach similar with `la.lisp` but I'm using MacPorts dependencies instead of `.tar`.

hebisch:
Dnia Mon, Jul 24, 2023 at 10:47:47AM -0700, Kirill A. Korinsky napisał(a):
> @hemmecke lisp images is like static linking of binary. I understand why it's happened but I'm pointing that it isn't good idea for complex system with a lot of dependencies.
> 
> FriCAS itself as far as I know hasn't got any lisp dependencies => here I may rationalise usage of kind of static linking. It doesn't change anything.
> 
> Well, it need to be rebuild after change of sbcl but it happens rarely.

I think there is some misunderstanding here.  When FriCAS is build
like distributed Linux binaries, then you can just put binaries on
target system and FriCAS is supposed to work.  No need to have
separate sbcl.  If you build from sources in default way you can
upgrade sbcl and for FriCAS nothing changes.

Concerning using executables during build, this was idea of Gabriel
Dos Reis (in Axiom era) and it considerably simplified build system.
I do not have numbers handy, but I have reasons to believe that
using executables makes FriCAS build go faster.

I understand that distributions work differently.  You want to
reuse what is present in the distribution and you take care
that everything is in matched versions.  Things are quite different
for binaries that FriCAS project distributes.  We do not have
manpower to create specific binaries for different distributions.
Instead, we create Linux binaries which have only "stable" dependencies,
that is dependencies which work in the same way on wide variety
of Linux systems.

Coming back to distributions: you probably have to live with using
executables during build (and consequently having executable as final
result).  Changing that is large effort and it is likely that effect
would be worse (slower and harder to maintain build system).

> Anyway, adding hunchentooth which has dozen dependencies makes things stranger.
> 
> >From my point of view exposing `SBCL_HOME` and/or `asdf` is the right way to support such dependencies. Quite universal and for any lisp which supports `asdf` and FriCAS.
> 
> But I agree that hunchentooth the first and probably the last dependencies of that kind.
> 
> To be clear: I've upgraded everything inside MacPorts, and removed my old patch to follow an approach similar with `la.lisp` but I'm using MacPorts dependencies instead of `.tar`.

Yes, the intent was that hunchentoot should be available, but hsbcl.tar
is just one of ways to provide it.

-- 
                              Waldek Hebisch

PR #122, opened 2023-05-02
GMP multiplication: simplify build process

gvanuxem:
Basically, this is a simplification of the build process in case GMP multiplication is used or not.

It adds in fricas-config.lisp a new LISP *features* :fricas_has_gmp and uses it to build or not num_gmp.lisp.
Additionally this features is used in fricas-lisp.lisp and util.lisp  to avoid the creation of two nil functions, init-gmp and uninstall-gmp-multiplication in num_gmpx.lisp, in case FriCAS is not built with GMP multiplication.


gvanuxem:
Sorry, GitHub has added all the history of my fork, my fault. Close it when/if necessary.

hebisch:
On Mon, May 01, 2023 at 08:30:00PM -0700, Greg wrote:
> Basically, this is a simplification of the build process in case GMP multiplication is used or not.
> 
> It adds in fricas-config.lisp a new LISP *features* :fricas_has_gmp and uses it to build or not num_gmp.lisp.
> Additionally this features is used in fricas-lisp.lisp and util.lisp  to avoid the creation of two nil functions, init-gmp and uninstall-gmp-multiplication in num_gmpx.lisp, in case FriCAS is not built with GMP multiplication.
> 
> 
>   https://github.com/fricas/fricas/pull/122

This really does not simplify.  Namely, we need 'num_gmpx.lisp' because
we use 'sed' to propagate information about path to GMP (usually we
just use 'libgmp.so', but some setups need different path).

Also, for me "simpler" means less conditionals.  That is why 'init-gmp'
and 'uninstall-gmp-multiplication' is always defined (but possibly
simply return nil).  I consider global flag (Lisp feature) more complicated
than this.

-- 
                              Waldek Hebisch

gvanuxem:
As a note question, I wonder why you're using 'sed' instead of #+:darwin && #-:darwin. I use this right now for a working copy of FriCAS.

hebisch:
Dnia Fri, Aug 25, 2023 at 09:47:57AM -0700, Greg napisał(a):
> As a note question, I wonder why you're using 'sed' instead of #+:darwin && #-:darwin. I use this right now for a working copy of FriCAS.

This is not a binary choice, 'sed' allows passing path to GMP from
'configure'.

-- 
                              Waldek Hebisch

PR #129, opened 2023-06-26
introduce *OBEY-STDOUT*

hemmecke:
This patch allows jFricas to catch (and show) certain output from FriCAS in a jupyter output cell, in particular system commands like ")system pwd".


hebisch:
As I wrote on the mailing list, this is OK.  Pleas commit.

-- 
                              Waldek Hebisch

PR #139, opened 2023-10-22
Fix collection and report of memory usage

oldk1331:
This patch will allow FriCAS to collect and report memory usage during computation.

Currently there is no toplevel settings to enable it, you shall use
````
)lisp (setq |$printStorageIfTrue| t)
````

For example:
````
(1) -> [x for x in 1..1000];

                                                  Type: List(PositiveInteger)
             Storage: 0 (IN) + 450,544 (EV) + 0 (OT) + 0 (GC) = 450,544 bytes
````

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2023-11-04  Qian Yun  <oldk1331@gmail.com>
+
+	* src/interp/g-timer.boot, src/interp/setvart.boot,
+	src/interp/macros.lisp, src/interp/setq.lisp:
+	Enable memory usage report: ')set message storage on'
+
 2023-10-30  Waldek Hebisch  <cas@fricas.org>
 
 	* src/lisp/Makefile.in: Use large memory model for GCL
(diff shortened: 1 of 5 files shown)


hebisch:
On Sat, Oct 21, 2023 at 05:33:49PM -0700, Qian Yun wrote:
> This patch will allow FriCAS to collect and report memory usage during computation.
> 
> Currently there is no toplevel settings to enable it, you shall use
> ````
> )lisp (setq |$printStorageIfTrue| t)
> ````

I think that we should have:

)set messages storage on

In other words, this functionality should be available without
using Lisp or Boot commands.

-- 
                              Waldek Hebisch

hemmecke:
Of course, and maybe also the shortcut ")storage on" in the same way 
that we already have ")time on" and ")type on".

Maybe next week I can find a reliable method to get the step number from 
FriCAS without relying on the "Storage: " line.
In fact, I only need a hook in FriCAS that outputs the stepnumber to 
some output stream from which jFriCAS can extract it, even the user has 
issued ")set message XXX off" and ")set output XXX off" for any 
available XXX.

Anyway even though I am not yet clear about this, I am in favour of
")set message storage on".

-- 
Ralf

Issue #142, opened 2023-11-27
SparseUnivariateLaurentSeries multiplies wrongly


hemmecke:
The following session returns
```
%%% (7) -> s2:=s*s

           2
   (7)  O(q )
                             Type: SparseUnivariateLaurentSeries(Integer,q,0)
%%% (8) -> [coefficient(s2,i) for i in 0..10]

   (8)  [0, 0, 39, 70, 49, 0, 0, 0, 0, 0, 0]
                                                          Type: List(Integer)
```
That is obviously wrong.

Changing to `)set stream calc 2` cures the result.
However, the correctness of the computation should not depend on the value of this global variable.

```
ZZ ==> Integer
SULS ==> SparseUnivariateLaurentSeries(ZZ, 'q, 0)
Rec ==> Record(k: ZZ, c: ZZ)
S ==> Stream Rec
)set stream calculate 1
st := cons([0,1]$Rec, cons([1,5]$Rec, cons([2,7]$Rec,empty()$S)))$S
s := series(st)$SULS;
s2:=s*s
[coefficient(s2,i) for i in 0..10]
```

hemmecke:
fixed by commit 5543cf52

PR #143, opened 2023-11-27
fix issue #142

hemmecke:
The problem is independent of ")set stream calc 1". It just happened to be hidden.

Before this commit the code below resulted in the values 0, 5, 3 for the last 3 commands.

ZZ ==> Integer
SULS ==> SparseUnivariateLaurentSeries(ZZ, 'q, 0)
Rec ==> Record(k: ZZ, c: ZZ)
S ==> Stream Rec
st := cons([0,3]$Rec, cons([1,5]$Rec, cons([2,7]$Rec,empty()$S)))$S s := series(st)$SULS;
i x ==> x pretend InnerSparseUnivariatePowerSeries ZZ explicitEntries?(getStream i s)
getRef i s
coefficient(s,0)
coefficient(s,1)
coefficient(s,0)


hebisch:
On Mon, Nov 27, 2023 at 01:43:50PM -0800, Ralf Hemmecke wrote:

> 
> https://github.com/fricas/fricas/pull/143.patch
> https://github.com/fricas/fricas/pull/143.diff

Change looks good.  But we should also add tests.

-- 
                              Waldek Hebisch

PR #147, opened 2023-12-10
update documentation for building FriCAS Aldor interface

hemmecke:
Unfortunately, this commit has some forgotten line from a commit correcting the documentation for jFriCAS. I can separate them (for INSTALL) if this is wanted, but INSTALL is generated anyway, so I did not bother.


hebisch:
On Sun, Dec 10, 2023 at 03:56:11PM -0800, Ralf Hemmecke wrote:
> Unfortunately, this commit has some forgotten line from a commit correcting the documentation for jFriCAS. I can separate them (for INSTALL) if this is wanted, but INSTALL is generated anyway, so I did not bother.

That is OK.

IIUC the

   ( cd src/aldor &&  make )

is no longer needed, '--enable-aldor' works.

-- 
                              Waldek Hebisch

PR #175, opened 2024-06-30
Fix Aldor interface generation to allow for % in constructor context

pbroadbery:
This allows ``Foo(X: KernelCategory %): with == add''

### DIFF
--- a/src/interp/ax.boot
+++ b/src/interp/ax.boot
@@ -157,7 +157,7 @@ modemapToAx(modemap) ==
       argtypes
   $modemapArgs : local := [cons(a,t) for a in args for t in atypes]
 
-  argdecls:=['Comma, : [axFormatDecl(a, stripType t) for a in args for t in argtypes]]
+  argdecls:=['Comma, : [axFormatDecl(a, stripType t, true) for a in args for t in argtypes]]
   resultType :=  axFormatType stripType target
   categoryForm? constructor =>
       categoryInfo := get_database(constructor, 'CONSTRUCTORCATEGORY)
@@ -204,8 +204,17 @@ optcomma [op,:args] ==
    # args = 1 => first args
    [op,:args]
 
-axFormatDecl(sym, type) ==
+-- flg indicates that the call is from constructor context
+axFormatDecl(sym, type, flg) ==
+   type is ['Join, :.] => ['Declare, sym, axFormatType type]
+   type is ['CATEGORY, :.] => ['Declare, sym, axFormatType type]
+   flg and containsSelf? type =>
+       ['Declare, sym, ['With, [], axFormatType type]]
    ['Declare, sym, axFormatType type]
+
+containsSelf? type ==
+    type is [op, :args] => or/[containsSelf? arg for arg in args]
+    type is '%
 
 makeTypeSequence l ==
    ['Sequence,: delete('Type, l)]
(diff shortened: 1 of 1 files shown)


hebisch:
On Sun, Jun 30, 2024 at 02:05:48PM -0700, pbroadbery wrote:
> This allows ``Foo(X: KernelCategory %): with == add&#39;&#39;
> 
>   https://github.com/fricas/fricas/pull/175
> 
> -- Commit Summary --
> 
>   * Fix Aldor interface generation to allow for % in constructor context

AFAICS this has potential trouble since '%' not always means
main constructor.  Namely in 'Foo(A : with f : % -> %) : B'
'%' means the same as 'A'.  Such usage is quite rare, so probably
the patch is good enough.  It allows building Aldor interface
and simple tests pass.  I did not run more complicated tests.

-- 
                              Waldek Hebisch

pbroadbery:
> AFAICS this has potential trouble since '%' not always means main constructor. Namely in 'Foo(A : with f : % -> %) : B' '%' means the same as 'A'. Such usage is quite rare, so probably the patch is good enough. It allows building Aldor interface and simple tests pass. I did not run more complicated tests.
> […](#)
> -- Waldek Hebisch

That's something that came up when writing it - The patch specifically deals with constructors "Foo(X: E(...))" where "E(...)" contains a %.  For the case you mention, the new 'flg' argument to axFormatDecl will be false. 
I do need to add a couple of smoke tests at the aldor level - but I think the fricas -> aldor side is fine

hemmecke:
> The patch specifically deals with constructors "Foo(X: E(...))" where "E(...)" contains a %.

Hmmm..., I think
```
Foo(X: E(%) with bar: %->%)
```
might make an interesting testcase.

The question for me is when exactly does % mean Foo and when does it mean X? Is it the same for FriCAS and Aldor?

hebisch:
On Mon, Jul 01, 2024 at 07:07:21AM -0700, pbroadbery wrote:
> 
> > AFAICS this has potential trouble since '%' not always means main constructor. Namely in 'Foo(A : with f : % -> %) : B' '%' means the same as 'A'. Such usage is quite rare, so probably the patch is good enough. It allows building Aldor interface and simple tests pass. I did not run more complicated tests.
> > […](#)
> > -- Waldek Hebisch
> 
> That's something that came up when writing it - The patch specifically deals with constructors "Foo(X: E(...))" where "E(...)" contains a %.  For the case you mention, the new 'flg' argument to axFormatDecl will be false.

I mean, 'flg' handles fine simple cases, but does not solve general case.
'containsSelf?' will by necessity recurse quite deep, but nested '%' may
by bound to different context.

As I wrote, this is probably good for now, but may came back later.
I fixed a few related problems in Spad compier, but AFAIK there are
still some cases which Spad compiler gets wrong.

Anyway, I think we should commit this now, as it solves current
problem.
-- 
                              Waldek Hebisch

hebisch:
On Mon, Jul 01, 2024 at 07:46:16AM -0700, Ralf Hemmecke wrote:
> > The patch specifically deals with constructors "Foo(X: E(...))" where "E(...)" contains a %.
> 
> Hmmm..., I think
> ```
> Foo(X: E(%) with bar: %->%)
> ```
> might make an interesting testcase.
> 
> The question for me is when exactly does % mean Foo and when does it mean X? Is it the same for FriCAS and Aldor?

AFAICS the rule is that embedded 'with' essentially creates new scope for
'%'.  I think that the rule is the same for FriCAS and Aldor, and that
deviations are unintentional.  OTOH due to simplistic implementation
Spad compiler not always follows the rule.

-- 
                              Waldek Hebisch

hemmecke:
Are you saying that in
```
Foo(X: with E(%) with bar: %->%)
```
the % inside the E(%) suddenly refers to X instead of Foo?

oldk1331:
Do we have an agreement to merge this commit in its current form?

hemmecke:
+1

oldk1331:
Merged as https://github.com/fricas/fricas/commit/2b4b65648ca095dd0379110899c83adadc62e39f, closing.

PR #180, opened 2024-08-25
fixing some typos

fchapoton:
just a few more typos ; please notice when merging

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1453,7 +1453,7 @@
 
 2022-07-09  Waldek Hebisch  <cas@fricas.math.uni.wroc.pl>
 
-	* src/algebra/mfinfact.spad: Fix missing trival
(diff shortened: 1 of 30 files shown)


fchapoton:
also the encoding of 2 files is fixed, which is maybe more controversial

oldk1331:
I have verified changes to `src/doc/ht/HTXFormatPage7.ht` cause display issues for HyperDoc.

Although not tried myself, but I believe changes to `src/hyper/lex.c` will cause problems.

I guess you changed the encoding to UTF-8?

If you revert those 2 files, I can merge the rest.

fchapoton:
here we go, without the UTF8 encoding changes in these two files

oldk1331:
Thanks!  Merged now.

PR #188, opened 2024-12-31
Update fricas-lisp.lisp: sb-alien::int => sb-alien::integer (sb-alien::int is a (signed-byte 32))

gvanuxem:
sb-alien::int was signalled by sbcl-2.5.0 on my FriCAS copy as a (signed-byte 32) whereas I was using most-positive-fixnum which is for sb-alien sb-alien:integer. That solved my issue, but of course if you are expecting a 32bit C int this proposal can be deleted.

### DIFF
--- a/src/lisp/fricas-lisp.lisp
+++ b/src/lisp/fricas-lisp.lisp
@@ -447,7 +447,7 @@ with this hack and will try to convince the GCL crowd to fix this.
 (eval-when (:compile-toplevel :load-toplevel :execute)
 
 (setf *c-type-to-ffi* '(
-    (int SB-ALIEN::int)
+    (int SB-ALIEN::integer)
     (c-string SB-ALIEN::c-string)
     (double SB-ALIEN::double)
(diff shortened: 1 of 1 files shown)


hebisch:
On Mon, Dec 30, 2024 at 06:56:41PM -0800, Grégory Vanuxem wrote:
> sb-alien::int was signalled by sbcl-2.5.0 on my FriCAS copy as a (signed-byte 32) whereas I was using most-positive-fixnum which is for sb-alien sb-alien:integer. That solved my issue, but of course if you are expecting a 32bit C int this proposal can be deleted.

'sb-alien::int' is used for calling C routines and should match C
types, so 'sb-alien:integer' would be wrong.

-- 
                              Waldek Hebisch

gvanuxem:
Yes, I saw, this is right, sorry for the noise. I use C long and for me it
was always a question about the int/long distinction on 32/64 bits
architecture. Digging into CL implementations documentation and C code used
by FriCAS let me better understand.

PR #193, opened 2025-03-19
Fix setchildren! not implemented for BinaryRecursiveAggregate

fahasch:
The BinaryRecursiveAggregate extends RecursiveAggregate. However setchildren! is not implemented (nor is it by any of the domains). The proposed fix is to have a default implementation that changes both children.

### DIFF
--- a/src/algebra/aggcat.spad
+++ b/src/algebra/aggcat.spad
@@ -1248,6 +1248,12 @@ BinaryRecursiveAggregate(S : Type) : Category == RecursiveAggregate S with
        setelt!(x, "left", b)  == setleft!(x, b)
        setelt!(x, "right", b) == setright!(x, b)
 
+       setchildren!(x, ls) ==
+           #ls ~= 2 => error "a binary tree needs 2 children"
+           setleft!(x, ls.1)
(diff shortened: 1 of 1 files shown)


oldk1331:
CI fails.  I can not explain why the type check does not pass.  But there is logic error in your code: you should check for the size to be less than 2:

````
       setchildren!(x, ls) ==
           less?(ls, 2) => error "a binary tree needs 2 children"
           setleft!(x, first ls)
           setright!(x, second ls)
           x
````

oldk1331:
Yes, I just realized that.  It should be `not size?(ls, 2)` .

hebisch:
On Wed, Mar 19, 2025 at 05:55:12PM -0700, Qian Yun wrote:
> oldk1331 left a comment (fricas/fricas#193)
> 
> Yes, I just realized that.  It should be `not size?(ls, 2)` .

Original was mostly OK.  The trouble is that compiler does
not use signatures from NonNegativeInteger.  This may be
compiler limitation.  Version below works:

           not(#ls =$NonNegativeInteger 2) =>
               error "a binary tree needs 2 children"

-- 
                              Waldek Hebisch

PR #196, opened 2025-04-13
introduce string join with separator

hemmecke:
I think such a function would be quite useful.

join(["a","b","cd"], " + ") returns "a + b + cd"

The question is whether it should live directly in IndexedString (or String) or whether we introduce a new package StringTools where (maybe) also other useful string functions can go.

### DIFF
--- a/src/algebra/aggcat.spad
+++ b/src/algebra/aggcat.spad
@@ -2597,6 +2597,10 @@ StringAggregate : Category == OneDimensionalArrayAggregate Character with
       ++ elt(s, t) returns the concatenation of s and t. It is provided to
       ++ allow juxtaposition of strings to work as concatenation.
       ++ For example, \spad{"smoo" "shed"} returns \spad{"smooshed"}.
+    join: (%, List %) -> %
+      ++ join(sep,ls) returns an element that is the concatenation
+      ++ of all strings in ls but separated by sep, for example,
+      ++ concat("/",["a","b","c"]) returns "a/b/c".
  add
    trim(s : %, cc : CharacterClass) == leftTrim(rightTrim(s, cc), cc)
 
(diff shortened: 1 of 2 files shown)


martinbaker:
Could this be made more general (not just Strings) by having another version of 'reduce' in List which inserts the separator?

hebisch:
On Mon, Apr 14, 2025 at 01:13:25AM -0700, martinbaker wrote:
> martinbaker left a comment (fricas/fricas#196)
> 
> Could this be made more general (not just Strings) by having another version of 'reduce' in List which inserts the separator?

One can use code like:

ff2(a : String, sep : String, l : List(String)) : List(String) ==
    empty?(l) => [a]
    cons(a, cons(sep, l))

reduce((s : String, l : List(String)) +-> ff2(s, " + ", l), ["a", "b", "c"], [])

which gives:

["c", " + ", "b", " + ", "a"]

reversing this (and concatenating if desired) gives the result.
In Spad code String above can be replaced by a domain parameter.

Of course, simple loop is of comparable size.

Since implementation is pretty easy users can add such thing as
needed.  OTOH I am not sure if there is a need, except for string
case.

-- 
                              Waldek Hebisch

hebisch:
On Sun, Apr 13, 2025 at 01:22:59PM -0700, Ralf Hemmecke wrote:
> I think such a function would be quite useful.
> 
> join([&quot;a&quot;,&quot;b&quot;,&quot;cd&quot;], &quot; + &quot;) returns &quot;a + b + cd&quot;
> 
> The question is whether it should live directly in IndexedString (or String) or whether we introduce a new package StringTools where (maybe) also other useful string functions can go.

Probably it is useful enough to include.  I think that implementation
could be somwhat simpler:

1) Something like:

    l := +/[#s for s in l] + #rest(l)*#sep

to compute length of the result


2) We should be able to reduce number of $Lisp calls by using helper
   function like

    copy_into(s1 : String, s2 : String, i : NNI) : NNI ==
        l := #s1
        RPLACSTR(s1, i, l, s2, 0, l)$Lisp
        l

We can call it from 'concat', 'copyInto!' and new 'join'.

Since this implementation of 'join' is closely related to 'concat'
it is natural to put it in the same file.

-- 
                              Waldek Hebisch

PR #197, opened 2025-05-18
Make fricas installation relocatable

mkoeppe:
Instead of hard-coding the `exec_prefix` in the installed `fricas` script, obtain it from the invocation of the script.

The portable shell code for this also handles the situation when the script is invoked via symlinks.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -154,6 +154,10 @@ install-src:
 	echo FRICAS_VERSION='"${PACKAGE_VERSION}"' >> '${COMMAND}'.tmp
 	echo FRICAS_LISP_FLAVOR='"${FRICAS_LISP_FLAVOR}"' >> '${COMMAND}'.tmp
 	echo FRICAS_LISP_VERSION='"${FRICAS_LISP_VERSION}"' >> '${COMMAND}'.tmp
+	if [ ${FRICAS_LISP_FLAVOR} = ecl -a -z "$$ECLDIR" ]; then \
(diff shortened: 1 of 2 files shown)


hebisch:
On Sun, May 18, 2025 at 10:40:47AM -0700, Matthias Köppe wrote:
> Instead of hard-coding the `exec_prefix` in the installed `fricas` script, obtain it from the invocation of the script.

I have some doubts about this patch.  First, current design allows
you to easily customize behaviour: you can prepend to 'fricas'
script your own script which computes desired FRICAS_PREFIX
(or set FRICAS_PREFIX in environment before invoking 'fricas'
script).  You remove this possibility of customization, replacing
it by fixed behaviour, that is depending on relative positions.

Second, FriCAS in principle runs on quite large variety of operating
systems.  My understanding was that some systems do not set "$0"
to full pathname.

Extra remark1: 'resolvelinks.sh' tries to produce relative path
when given a relative path as input.  But the only supported
seting of FRICAS is absolute.

Extra remark2: FRICAS_PREFIX exists to support installers on
Mac OSX.  I do not know if it is used in current packages,
but if used your change can in principle break them.

-- 
                              Waldek Hebisch

mkoeppe:
> FRICAS_PREFIX exists to support installers on Mac OSX. I do not know if it is used in current packages, but if used your change can in principle break them.

Likely it is used exactly to relocate the prebuilt binary package, and they would simply not have to set this variable any more.

mkoeppe:
> FriCAS in principle runs on quite large variety of operating systems. My understanding was that some systems do not set "$0" to full pathname.

I can only note that the script was in use in Sage for a very long time, including when Cygwin was still supported. The script is so long because it achieves its goals even when "$0" is not the full pathname.

PR #220, opened 2026-07-28
WM_DELETE_WINDOW support for hypertex

tkorvola:
Makes close buttons provided by the window manager work properly. Some WMs kill clients instead if they don't support the protocol.

### DIFF
--- a/src/hyper/event.c
+++ b/src/hyper/event.c
@@ -180,6 +180,11 @@ handle_event(XEvent * event)
       case DestroyNotify:
 /*        fprintf(stderr,"event:handle_event type=DestroyNotify\n");*/
         break;
+      case ClientMessage:
+        if (event->xclient.message_type == gxa_wm_protocols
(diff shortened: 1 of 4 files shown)


oldk1331:
Hi, can you share which WM is affected by this?

tkorvola:
The WM_DELETE_WINDOW protocol is standard (ICCCM Section 4.2.8.1). Window manager support should be quite universal now, starting with TWM. Clients declare support per window. What happens when the user tries to close a window that does not support the protocol, is not standard. KWin (on X or Wayland) kills the client. If I recall correctly, Sawfish does the same with default settings.

oldk1331:
Thanks, just tried it out, the error message `X connection to :0 broken (explicit kill or server shutdown).` is gone when closing hyperdoc window with "x" button in title bar instead of the "Exit" button in hyperdoc.

I'll discuss with other developers in mail list to include this commit.

Issue #221, opened 2026-09-01
src/doc: "sman -paste" sometimes never returns, hanging the build


ilgrad:
Building master 46f4240 on x86-64 Fedora 44 with SBCL 2.6.8, the build stops
inside src/doc while generating the HyperDoc viewports, and stays stopped. It
does not fail: make simply never gets its child back.

Frequency here: 5 of 14 builds in a clean chroot, always on the same file,
ug01.ht. An idle machine and a loaded one both do it.

State when it is stuck. The last line of output is the final page of that file,

    parsing: IFSNodePage

and the processes are

    sh /usr/bin/xvfb-run -a -s '-screen 0 1024x768x24' .../sman -noihere -paste ug01.ht
    .../sman -noihere -paste ug01.ht
    .../bin/hypertex -s -k -ip ug01.ht
    .../lib/viewman
    .../sman -noihere -paste ug01.ht

All of them in state S at 0% CPU, indefinitely - I have left it for 36 minutes.
FRICASsys has already exited by then, so the work itself is finished and what
is left is the shutdown. The .VIEW output that the successful runs go on to
copy is complete.

A successful run continues from that same point with

    cp -r *.VIEW .../share/hypertex/viewports
    echo timestamp > viewports

so the difference is only whether sman, hypertex and viewman notice they are
done.

Not a parallel make problem, though there is one nearby. I first assumed the
cause was several sman instances at once: src/doc/Makefile.in:499 is a pattern
rule over ${GEN_VIEWPORTFILES_PHT}, sman serves its sockets under the fixed
names /tmp/.i<n> and /tmp/.h<n> (src/include/com.h:107) and sweeps /tmp for
stale ones on startup (src/sman/sman.c:372), so two at a time are not safe.
Adding

    .NOTPARALLEL: ${GEN_VIEWPORTFILES_PHT}

does serialise them - verified, one .ht at a time in the process list - and the
hang still happens. So the two are separate, and this report is about the
second one. I am happy to send the .NOTPARALLEL change as a pull request on its
own if you want it; it is correct regardless.

What I am doing meanwhile, in case it is useful to someone else hitting this: I
bound each attempt and retry it, which turns an unbounded stall into about
three extra minutes.

    ok=no
    for try in 1 2 3; do
        if FRICAS_INITFILE='' timeout 180 ${XVFB} ${SMAN} -noihere -paste $<; then ok=yes; break; fi
        echo "viewport generation for $< did not finish, attempt $try"
    done
[...]

PR #222, opened 2026-09-01
doc: do not build the viewports in parallel

ilgrad:
Each `GEN_VIEWPORTFILES_PHT` target runs its own `sman`. `sman` serves its
sockets under the fixed names `/tmp/.i<n>` and `/tmp/.h<n>`
(`src/include/com.h:107`) and sweeps `/tmp` for stale ones when it starts
(`src/sman/sman.c:372`), so two of them running at the same time can unlink
each other's live sockets.

A parallel make does start two: on this machine the process list showed
`-paste ug01.ht` and `-paste ug10.ht` running together, each with its own
`hypertex` and `viewman`.

GNU make 4.4 and later apply `.NOTPARALLEL` with prerequisites to just those
targets, so the rest of `src/doc` still builds in parallel; older make
serialises the whole file, which is slower but also correct.

Verified on x86-64 Fedora 44, GNU make 4.4.1, SBCL 2.6.8: with this change the
process list never holds more than one `.ht` at a time, and `make check`
reports `no unexpected failures` across five consecutive clean-chroot builds.

This is not a fix for #221 — I checked, the hang described there still happens
with the viewports serialised — but the two are independent and this one stands
on its own.

### DIFF
--- a/src/doc/Makefile.in
+++ b/src/doc/Makefile.in
@@ -320,6 +320,12 @@ VIEWPORTNAMESviewports = ${VIEWPORTFILES}
 # files in ${GEN_VIEWPORTFILES_PHT} involve the generation of
 # graphics, the files in ${GEN_NONVIEWPORTFILES_PHT} do not.
 GEN_VIEWPORTFILES_PHT = ${patsubst %, %.pht, ${VIEWPORTNAMES${MAYBE_VIEWPORTS}}}
+
+# Each of these runs its own sman, which serves sockets under the fixed
+# names /tmp/.i<n> and /tmp/.h<n> and sweeps /tmp for stale ones at startup,
+# so two at once can tread on each other.  GNU make 4.4 and later restrict
+# this to the named targets, older make serialises the whole file.
+.NOTPARALLEL: ${GEN_VIEWPORTFILES_PHT}
 GEN_NONVIEWPORTFILES_PHT = ${patsubst %, %.pht, \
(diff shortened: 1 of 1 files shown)


hebisch:
On Tue, Sep 01, 2026 at 03:47:37AM -0700, Ilia Gradina wrote:
> Each `GEN_VIEWPORTFILES_PHT` target runs its own `sman`. `sman` serves its
> sockets under the fixed names `/tmp/.i<n>` and `/tmp/.h<n>`
> (`src/include/com.h:107`)

They differ in process id which is appended to the base name.

> and sweeps `/tmp` for stale ones when it starts
> (`src/sman/sman.c:372`), so two of them running at the same time can unlink
> each other's live sockets.

'check_spad_proc' checks if socket is live and unlinks only stale
ones.

You can start several copies of FriCAS and closing one of them
should close stale sockets, but live one should stay.  And everday
practice shows that this works.

So this change is not needed.

-- 
                              Waldek Hebisch

ilgrad:
You are right on both points, and I should have followed the call chain instead
of guessing at it. `make_server_number` is `getpid()`
(`src/lib/sockio-c.c:851`), so `/tmp/.i<n>` and `/tmp/.h<n>` are unique per
process by construction, and `check_spad_proc` unlinks only after
`kill(pid, 0)` returns `ESRCH` (`src/sman/sman.c:361`), so a live sibling's
socket is never touched. My premise was wrong.

Closing. Sorry for the noise. The hang in #221 is separate and does not depend
on this.

